PR #5, opened 2018-01-22
DID class and related tests, fixed package.json testing script

### LINKED ISSUE #2: Define DID class in accordance with DID specifications v1
Did should be a colon separated string. It consists of:
- `did`
- `jol` implying that it's a Jolocom identity
-  a public key based string which ensures uniqueness of did

charleenfei:
Fixes #2


ts/internal_modules/identity.ts
```diff
@@ -0,0 +1,9 @@
+import * as bs58 from 'bs58'
+
+export default class DID {
+  identifier: string
+  constructor(publicKey: Buffer) {
+    const fragment = bs58.encode(publicKey)
```
katszwn: I would rename it - it's actually a whole, encoded public key, not a fragment.

ts/internal_modules/identity.ts
```diff
@@ -0,0 +1,9 @@
...
+    this.identifier = 'did:jolo:'+ fragment
```
katszwn: I would move `did:jolo` to a constant inside the class definition

js/internal_modules/identity.d.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference types="node" />
+export default class DID {
```
katszwn: We need some general docs here and perhaps a link to DID specification

PR #10, opened 2018-01-24
Created key generation module

### LINKED ISSUE #4: Move key generation from the smartwallet
- [x] Extract master key pair and generic signing key derivation to a separate module. 

- [x] Testing coverage

charleenfei:
Fixes #4


Exulansis:
@katszwn I agree, if too much test data starts piling up, we have to move it to a separate folder, to have a cleaner distinction.
@charleenfei Please consider adding a label on this (i.e. in progress or review required), and reference the issue it is related to (i'd assume #4). Thanks!

tests/key-derivation.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import testData from './data/key-data'
+import * as bitcoin from 'bitcoinjs-lib'
+import * as bip39 from 'bip39'
+
+
+describe('deriveMasterKeyPairFromSeedPhrase', () => {
+  it('Should correctly derive a master key pair given a seedphrase', () => {
+    const seedphrase = testData.testSeedPhrase
+    const result = deriveMasterKeyPairFromSeedPhrase(seedphrase)
+    expect(JSON.stringify(result)).to.equal(testData.testMasterKeyPairString)
```
katszwn: In this scenario we should also check equality of the notation of keys we'll be encrypting - which is the `WIF` format I think.

tests/data/key-data.ts
```diff
@@ -0,0 +1,11 @@
...
+  testGenericKeyPairString: '{"keyPair":{"d":{"0":29420278,"1":48403266,"2":32085146,"3":39027795,"4":56679644,"5":37054444,"6":34166252,"7":27070958,"8":8576908,"9":4007242,"10":0,"t":10,"s":0},"compressed":true,"network":{"messagePrefix":"\\u0018Bitcoin Signed Message:\\n","bech32":"bc","bip32":{"public":76067358,"private":76066276},"pubKeyHash":0,"scriptHash":5,"wif":128}},"chainCode":{"type":"Buffer","data":[79,83,158,85,219,181,148,72,160,10,6,60,177,234,246,132,223,16,204,117,236,7,56,25,47,211,161,129,194,30,157,82]},"depth":3,"index":2147483648,"parentFingerprint":1206978215}',
+  testEthereumKeyPairString: '{"keyPair":{"d":{"0":34429630,"1":29843365,"2":50470156,"3":55799221,"4":8282470,"5":47268858,"6":58968987,"7":16414218,"8":66677574,"9":2817547,"10":0,"t":10,"s":0},"compressed":true,"network":{"messagePrefix":"\\u0018Bitcoin Signed Message:\\n","bech32":"bc","bip32":{"public":76067358,"private":76066276},"pubKeyHash":0,"scriptHash":5,"wif":128}},"chainCode":{"type":"Buffer","data":[91,187,73,205,247,91,60,73,235,223,80,179,76,148,177,39,125,183,96,24,245,65,243,229,174,144,181,82,147,164,91,23]},"depth":5,"index":0,"parentFingerprint":3464247385}'
```
katszwn: We should also put different format of the keys here - the one we'll be using.

tests/key-derivation.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  })
+})
+
+
+describe('deriveGenericSigningKeyPair', () => {
+  it('Should correctly derive a generic signing key from a master key pair', () => {
+    const masterKeyPair = new bitcoin.HDNode.fromSeedBuffer(testData.testSeed)
+    const result =  deriveGenericSigningKeyPair(masterKeyPair)
+    expect(JSON.stringify(result)).to.equal(testData.testGenericKeyPairString) })
```
katszwn: The same here

tests/key-derivation.test.ts
```diff
@@ -0,0 +1,31 @@
...
+})
+
+
+describe('deriveEthereumKeyPair', () => {
+  it('Should correctly derive an Ethereum key pair from a master key pair', () => {
+    const masterKeyPair = new bitcoin.HDNode.fromSeedBuffer(testData.testSeed)
+    const result =  deriveEthereumKeyPair(masterKeyPair)
+    expect(JSON.stringify(result)).to.equal(testData.testEthereumKeyPairString)
```
katszwn: And here :)

ts/identity/key-derivation.ts
```diff
@@ -0,0 +1,39 @@
...
+/* @summary - Generate a generic signing key according to BIP32 specification
+ *
+ * @param {HDNode} masterKeyPair - the master keypair used to
+ * derive the child key.
+ *
+ * @returns {HDNode} - the derived child  generic signing key.
+ *
+*/
+
+export function deriveGenericSigningKeyPair(masterKeyPair: any) {
```
katszwn: I think it actually needs an `HDNode` instance, otherwise it will throw an exception. We should avoid using `any` as a type - the type acts as a first line of validation of data we get in a method.

ts/identity/key-derivation.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as bip39 from 'bip39'
+
+// @summary - Generates a keypair based on provided entropy
+//eslint-disable-next-line
+// @param {String} seedPhrase - a BIP39 compliant mnemonic generated from hashed entropy.
+//eslint-disable-next-line
+// @returns {HDNode} - an instance containing a master keypair and a default Bitcoin network object.
+
+
+export function deriveMasterKeyPairFromSeedPhrase(seedPhrase: string) {
```
katszwn: We should define the return type here, for readability. The same in methods below

charleenfei: *NOTE*: type-guarding to be implemented with the Keypair class implementation

ts/identity/key-derivation.ts
charleenfei: *NOTE*: type-guarding to be implemented with the Keypair class implementation

tests/key-derivation.test.ts
charleenfei: fixed in next commit

PR #11, opened 2018-01-24
Add DID Document class, handle serialization

### LINKED ISSUE #1: Define DID Document class in accordance with DID/DDO specifications
## Description
It should contain:
- `@context` - the version of DID specifications
- `id` unique identifier with colon separated description, ex. `did:jol:2829fen32992jkjdk2o3`
- `authenticationCredential` a separate class, describing public key that is used by the owner 
- `signature` a JSON-LD compatible signature, example [here](https://w3c-dvcg.github.io/ld-signatures/)
- `created` timestamp of identity creation

## TODO
- [x] Add DidDocument class
- [x] Add AuthorizatioCredential class
- [x] de- and serialization of AuthorizationCredential class
- [x] de- and serialization of DidDocument class

katszwn:
Fixes #1 

Adds data mode for DID Document, serializes it to W3C compliant format, deserializes back to corresponding Typecript object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   "dependencies": {
     "bip39": "^2.5.0",
     "bitcoinjs-lib": "^3.3.2",
-    "bs58": "^4.0.1",
     "colors": "^1.1.2",
     "istanbul": "^0.4.5",
     "web3": "^1.0.0-beta.29",
--- a/package.json
+++ b/package.json
@@ -37,8 +37,9 @@
     "bs58": "^4.0.1",
     "colors": "^1.1.2",
     "istanbul": "^0.4.5",
-    "nyc": "^11.4.1",
-    "web3": "^1.0.0-beta.29"
+    "web3": "^1.0.0-beta.29",
+    "lolex": "^2.3.1",
+    "nyc": "^11.4.1"
   },
   "devDependencies": {
     "@types/mocha": "^2.2.43",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -37,8 +37,9 @@
     "bs58": "^4.0.1",
```
Exulansis: We don't need this anymore, right?

ts/identity/did.ts
```diff
@@ -1,24 +1,25 @@
-import { utils } from 'web3'
+import * as bs58 from "bs58";
```
Exulansis: It hash been changed to sha3.

ts/identity/did.ts
```diff
@@ -1,24 +1,25 @@
...
+    });
   }
 
-  toJSON(): string {
-    return this.identifier
+  public identifier: string;
+
+  constructor(publicKey: Buffer) {
+    const prefix = "did:jolo:";
+    const suffix = bs58.encode(publicKey.subarray(0, 16));
```
Exulansis: Needs to be updated, this was changed on master I believe.

ts/tsconfig.json
```diff
@@ -1,20 +1,20 @@
 {
```
Exulansis: Poor file.

ts/identity/did.ts
Exulansis: Namely the suffix is now generated using the sha3 function

package.json
Exulansis: This was only used for generating the DID suffix.

PR #12, opened 2018-01-26
Authentication procedure

### LINKED ISSUE #3: Add a basic authentication procedure
For authentication, we should use a mechanism of cryptographic signature with a private key and verification of the signature with a public key. We can use https://github.com/digitalbazaar/jsonld-signatures to create JSON-LD compatible signature.

nataschaberg:
Fixes #3


ts/sso/authentication.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    return tokenData.payload.claims
+  } else {
+    return new Error('Web Token Not Valid')
+  }
+}
+
+
+function _getSigningKeysFromWIF({WIF} : {WIF : string}) {
```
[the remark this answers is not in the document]
katszwn: This particular method would go to a generic helpers file, perhaps `ts/utils/keysEncoding.ts`, or even just one file for all helper functions, `ts/utils.ts` - WDYT?

ts/sso/tokenPayload.ts
```diff
@@ -0,0 +1,50 @@
+export default class TokenPayload {
```
katszwn: We should add some docs here

ts/security/sign-verify.ts
```diff
@@ -0,0 +1,20 @@
+import * as bitcoinMessage from 'bitcoinjs-message'
+import * as bitcoin from 'bitcoinjs-lib'
+import * as bip39 from 'bip39'
+
+export function signMessage(WIF : String, message : String) : String {
+  const keyPair = bitcoin.ECPair.fromWIF(WIF)
+  const privateKey = keyPair.d.toBuffer(32)
+
+  const sigBuffer = bitcoinMessage.sign(message, privateKey, keyPair.compressed)
```
katszwn: `signBuffer`?

ts/sso/tokenPayload.ts
katszwn: Also, 👍 for extracting it as a class! 🎉

ts/sso/authentication.ts
```diff
@@ -0,0 +1,145 @@
...
+import {
+  initiateSecretExchange,
+  respondSecretExchange,
+  getEncryptionSecret,
+  encrypt,
+  decrypt
+} from '../security/encryption'
+
+
+export async function initiateRequest({did, claims, IPFSroom, WIF, encrypt} :
```
katszwn: This could be a class named something like `AuthenticationAgent`

PR #14, opened 2018-01-26
Integrate IPFS storage

### LINKED ISSUE #13: Integrate IPFS storage
### Description

This library should ideally be storage agnostic, different parts can be imported by the client application to create and validate DDOs. Since different systems might have different storage possibilities for the underlying claims / DDOs, we provide basic methods for creating and validating / verifying them. 
For applications that do not want to handle storage themselves, a basic default solution on top of IPFS will be provided.

### TODO:

- [x] Implementation of storage.
- [x] Research / Implementation of IPNS support.
> Currently IPNS resolutions are slow (<120 sec), there seems to be no way of delegating IPNS name registration to public IPFS gateways without them owning the controlling key-pair.

> An alternative approach using IPFS pub sub was explored, it does not require the user to give up control of any keys, but it does require a constant service to run, that subscribes to the appropriate topics and notifies interested parties. Still under considerations.

> A simple approach using one Ethereum smart contracts will most likely be implemented for now [#15]. The associated running costs will have to be assessed. This is most likely not the final solution. 

- [ ] Set up permissioned IPFS gateway.
- [x] Test coverage.

Exulansis:
Fixes #13


Exulansis:
Currently depends on #1. After the #1 is merged into master, this can be reviewed and merged.

PR #25, opened 2018-02-06
Verifiable Credential interface, class, test

### LINKED ISSUE #23: Create and update public verifiable credentials
### Description

The library should expose methods to create, sign, and verify claims about the identity.

### TODO:

- [x] Define verifiable credential structure.
- JWT v. JSON-LD: JWT is neater with less convoluted (and/or irrelevant) syntax. But, JWT is a base64 encoded blob -- you can't see the content unless you decode it, which is potentially a problem or limitation for potential [searchable public credentials](https://github.com/ipfs-search/ipfs-search)
- Consider JWT for 'private' claims (ES256 alg in header for asymmetric decoding) and JSON-LD for public?
- JSON-LD VC structure would follow the W3C Verifiable Claims standard, and use a [W3C signature suite](https://github.com/digitalbazaar/jsonld-signatures) for a LD signature.
- Which one could best be expanded to handle multisig situations? uPort goes with array of JWT to represent multi-sig. Possibility: set IPFS claim hashes into IPLD as [merkle-dags](https://github.com/ipld/js-ipld-graph-builder) for multisig?
- [ ] Refactor DID class to include method for creating, signing, and verifying verifiable credential.
- typechecking in method for valid DID?
- [x] Add DID interface

charleenfei:
Fixes #23


tests/data/identity.ts
```diff
@@ -26,5 +26,17 @@ export default {
...
+    id: '11111',
+    'type': ["Credential"],
+		issuer: 'did:jolo:0x0aecaae09eb4b6433a4136fbac29a5ea93dd3593dd120e16a314744d3945d119',
+ 		issued:'Wed Jan 24 2018 16:42:15 GMT+0100 (CET)',
+    claim: {
+      id:'did:jolo:0xd0ae58da9f72c48767b04f339a1a0142bb8e86b521d008ca65f7e3983b03d32b',
+      ageOver: 21
+    }
+  },
+  expectedSignedCredential: '{"credential":{"@context":"https://w3id.org/credentials/v1","id":"11111","type":["Credential"],"issuer":"did:jolo:0xd0ae58da9f72c48767b04f339a1a0142bb8e86b521d008ca65f7e3983b03d32b","issued":"Wed Jan 24 2018 16:42:15 GMT+0100 (CET)","claim":{"id":"did:jolo:0xd0ae58da9f72c48767b04f339a1a0142bb8e86b521d008ca65f7e3983b03d32b","ageOver":21}},"signature":"a67544cb997e08f3c01791b5cf34f44c46244b5c6d1d45de4bbb5a60e249028e5a7337f65ddac10c7365e5f7652fb73cc8eddc687d9c6487b76f3d41013842fd"}'
```
katszwn: Maybe we could break lines to make it look like other test object, one attribute per line?

tests/did.test.ts
```diff
@@ -1,12 +1,39 @@
...
-import testData from './data/identity'
+import testDataIdentity from './data/identity'
+import testDataKeys from './data/keys'
+import * as bitcoin from 'bitcoinjs-lib'
+import * as lolex from 'lolex'
+import * as sinon from 'sinon'
 
 describe('DID', () => {
+
+    const publicKey = Buffer.from(testDataIdentity.testUserPublicKey, 'utf8')
```
katszwn: Shouldn't we use `hex` as encoding? From what I was testing, when we generate a hex from a public key Buffer, and then create a Buffer out of hex, we get a wrong Buffer that encodes to a different hex

tests/did.test.ts
```diff
@@ -1,12 +1,39 @@
...
-    const result = testData.testUserDID
+    const expected = testDataIdentity.testUserDID
     expect(did.identifier).to.be.a('string')
-    expect(did.identifier).to.equal(result)
+    expect(did.identifier).to.equal(expected)
+  })
+
+
+  it('Should contain a method which correctly generates and returns a signed verifiable credential' , () => {
+		const credentialType = ['Credential']
```
katszwn: Adjustment is wrong here - you can use visual mode in Vim, select these lines and use `=` to fix it :)

tests/did.test.ts
```diff
@@ -1,12 +1,39 @@
...
+    const result = did.createVerifiableCredential(credentialType, claim, testDataKeys.testGenericKeyPairWIF)
+    expect(JSON.stringify(result)).to.equal(testDataIdentity.expectedSignedCredential)
+  })
+
+  it('Should contain a method which can verify a signed credential', () => {
+    const genericKeyPair = bitcoin.ECPair.fromWIF(testDataKeys.testGenericKeyPairWIF)
+    const genericPublicKey = genericKeyPair.getPublicKeyBuffer()
+
+		const credentialType = ['Credential']
+		const claim ={id:'did:jolo:0xd0ae58da9f72c48767b04f339a1a0142bb8e86b521d008ca65f7e3983b03d32b', ageOver:21}
```
katszwn: And here

tests/verifiableCredential.test.ts
```diff
@@ -0,0 +1,18 @@
...
+import VerifiableCredential from '../ts/identity/verifiableCredential'
+import testData from './data/identity'
+import * as lolex from 'lolex'
+import * as sinon from 'sinon'
+
+describe('VerifiableCredential', () => {
+  const clock = lolex.install({now: new Date('2018-01-24T15:42:15.882Z')})
+  sinon.stub(VerifiableCredential.prototype, '_generateVerifiableCredentialID').returns('11111')
+
+	it('Should correctly instantiate a Verifiable Credential', () => {
```
katszwn: Adjustment issue here as well (and it will probably look fine in Vim, mine does it sometimes too - will need to look into it)

ts/identity/verifiableCredential.ts
```diff
@@ -0,0 +1,29 @@
+import { IVerifiableCredentialAttrs } from './types'
+import * as sjcl from 'sjcl'
+
+export default class VerifiableCredential implements IVerifiableCredentialAttrs {
+  public '@context': string = "https://w3id.org/credentials/v1"
+	public id: string
+  public 'type': string[]
+	public issuer: string
```
katszwn: And here

ts/identity/verifiableCredential.ts
```diff
@@ -0,0 +1,29 @@
...
+
+	constructor(credentialType: string[], issuerID: string, claim: { id: string; [x:string]:any }) {
+		this.id =  (this as any)._generateVerifiableCredentialID()
+		this.type = credentialType
+		this.issuer = issuerID
+		this.issued = (this as any)._dateIssued()
+		this.claim = claim
+  }
+
+	private _generateVerifiableCredentialID() : string {
```
katszwn: If it's marked as private, we don't need to use underscore in the name

ts/identity/verifiableCredential.ts
```diff
@@ -0,0 +1,29 @@
+import { IVerifiableCredentialAttrs } from './types'
+import * as sjcl from 'sjcl'
+
+export default class VerifiableCredential implements IVerifiableCredentialAttrs {
+  public '@context': string = "https://w3id.org/credentials/v1"
+	public id: string
+  public 'type': string[]
+	public issuer: string
+	public issued: string
+  public claim: { id: string; [x:string]:any }
```
katszwn: We could model `Claim` as a separate class, just like we did with DDO. But we can also extract it later, not necessarily in this PR. If you think it's a good idea, please open an issue.

ts/utils/utils.ts
```diff
@@ -0,0 +1,24 @@
+import * as createHash from 'create-hash'
+import * as secp256k1 from 'secp256k1'
+
```
katszwn: I think we could split it into separate files - one where sha256 and hash256 could go is the one introduced by @nataschaberg - `utils/keysEncoding` (perhaps a more general name, encoding could be used), and the other could be `utils/signature`.

tests/did.test.ts
charleenfei: fixing this :)

ts/identity/verifiableCredential.ts
charleenfei: I did think about this as well, we can do it in a separate PR 👍

ts/identity/verifiableCredential.ts
charleenfei: fixed!

tests/did.test.ts
charleenfei: fixed!

tests/did.test.ts
charleenfei: fixed!

PR #30, opened 2018-02-08
Include authentication in lib interface

### LINKED ISSUE #28: Include Authentication in the library interface
### Description
Expose Authentication module in the library

### TODO
- [x] create a class Authentication that encapsulates all needed functionalities
- [x] create authentication config 
- [x] include Authentication in Jolocom lib

katszwn:
Fixes #28


tests/authentication.test.ts
```diff
@@ -1,34 +1,12 @@
 import { expect } from 'chai'
 import testSignVerifyData from './data/sign-verify'
 import testAuth from './data/authentication'
+import Authentication from '../ts/sso/index'
 import * as QRCode from 'qrcode'
 import * as bitcoin from 'bitcoinjs-lib'
 import { TokenSigner, TokenVerifier, decodeToken } from 'jsontokens'
-import { initiateRequest, authenticateRequest, initiateResponse, authenticateResponse} from './../ts/sso/authentication'
 
-describe('create QR Code', () => {
```
katszwn: I removed this test as it was not using any of our code, just testing the functionality of `qrcode` library. Also, you added another test which checks the url - which takes a while, so we shouldn't duplicate it.

nataschaberg: good, thanks!

ts/sso/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  private async createQRCode({token} : {token: string}) : Promise<any> {
+    return new Promise((resolve, reject) => {
+      QRCode.toDataURL(token, (err, url) => {
+        if(err) { reject(err) }
+        resolve(url)
+      })
+    })
+  }
+}
```
nataschaberg: OK, cool. I think having it as a class is much cleaner.

tests/authentication.test.ts
```diff
@@ -117,49 +94,48 @@ describe('authentication process SSO', () => {
...
     expect(res).to.be.an('error')
   })
 
   it('initiateResponse should return a token ', (done) => {
-    initiateResponse({
+    authentication.initiateResponse({
       tokenData: tokenDataReq,
       WIF: testAuth.WIF,
       did: testAuth.mockDIDSUB,
       claims: [{name: 'Natascha'}, {hobby: 'day trading'}]
```
charleenfei: a little admission here!

Issue #32, opened 2018-02-10
Typescript error


Exulansis:
### Description

The way we are currently initializing our [here]() causes typescript to error out. An alternative way of populating a buffer with the random bytes received needs to be implemented.

### TODO
- [x] Implement alternative method
- [x] Assure it produces same hex string

PR #33, opened 2018-02-10
Changed the way we are writing the bytes to prevent TS errors.

Exulansis:
Addresses #32.

### DIFF
--- a/ts/identity/verifiableCredential.ts
+++ b/ts/identity/verifiableCredential.ts
@@ -9,21 +9,24 @@ export default class VerifiableCredential implements IVerifiableCredentialAttrs 
   public issued: string
   public claim: { id: string; [x:string]:any }
 
-	constructor(credentialType: string[], issuerID: string, claim: { id: string; [x:string]:any }) {
-		this.id =  (this as any).generateVerifiableCredentialID()
-		this.type = credentialType
-		this.issuer = issuerID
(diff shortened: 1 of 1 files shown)


Exulansis:
Would be nice to pass the size of the random string as a function argument, having values hardcoded is not too flexible. We need to think of a good way to do it, since entropy is given in 4 byte increments. If the return value is hex encoded, the smallest current output (1 word) is 8 hex characters and can be increased in increments of 8, which does not provide too much flexibility.

Perhaps this is not too important though.

ts/identity/verifiableCredential.ts
```diff
@@ -9,21 +9,24 @@ export default class VerifiableCredential implements IVerifiableCredentialAttrs
...
 
-	private generateVerifiableCredentialID() : string {
-		const UintArray = new Uint32Array(sjcl.random.randomWords(2))
-    const buf = Buffer.from(UintArray.buffer)
-    return buf.toString('hex')
-	}
+  private generateVerifiableCredentialID() : string {
+    const result = Buffer.allocUnsafe(8)
+    sjcl.random.randomWords(2).forEach((el, index) => {
+      result.writeInt32LE(el, index * 4)
```
katszwn: Niiiice!! Didn't think about it 👍

PR #38, opened 2018-02-14
expiry date implementation for verified claims

nataschaberg:


nataschaberg:
Please disregard the failed check; it is refering to index and verifiableCredential files having the same inputs.

Issue #26, opened 2018-02-07
[WIP] Store and retrieve claims on IPLD


charleenfei:
### Description

The library should expose API calls to store and retrieve verified credentials on IPFS, update the DDO with the relevant IPFS endpoints, and update the Registry Contract with the new DID/DDO mappings.

### TODO

- [x] Store and retrieve credentials on IPLD.
- add functionality in constructor for [creating DAGNode head](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DAG.md#dagput).
- store returned multihash as endpoint in DDO.
- create [CID](https://github.com/ipld/js-cid) from base58 encoding of multihash to retrieve node for later usage.
- create DAGNode for new Credentials object, convert this DAGNode to DAGLink.
- [add DAGLink](https://github.com/ipld/js-ipld-dag-pb#create-a-dagnode) to head DAGNode, accessible via the name(claimID) of the link (ie: headDAGNodeHash/claimID).
- new headDAGNode hash is generated, update DDO to reflect this new hash.
- update Registry Contract record with new DID/DDO mapping.
- [x] Add library method for storing publicly accessible credentials on IPLD which also updates the DDO and the Registry Contract record with the new DID/DDO mapping.
- Update DDO object to include credentials endpoints object.
- Update DDO to reflect new endpoint.
- Update Registry Contract record with new mapping.
- [x] Add library method for retrieving publicly accessible credentials on IPLD.
- [ ] Refactor to DAG API once ready.

Exulansis:
Closing for now, will reopen when this becomes an issue.

PR #39, opened 2018-02-21
Store and retrieve verifiable claims through IPLD

charleenfei:
Addresses #26


Exulansis:
This will be closed for now for housekeeping purposes. As soon as this issue becomes relevant again we can revisit this.

Issue #56, opened 2018-05-28
Allow for looking up identities by DID


Exulansis:
### Description

A method for looking up identities according to their DID should be available. Ideally it should return an object with useful methods exposed. For now, in the interest of time, the DDO as a JSON object should be returned. Further improvements can be added later.

### TODO
- [x] Define method for looking up identities according to their DID

PR #57, opened 2018-05-28
Added a basic identity resolution method.

Exulansis:
Addresses #56 .
As per the introduced TODO statement, changes to some initialization logic will need to be performed eventually.


ts/identity/index.ts
```diff
@@ -50,8 +51,9 @@ export class Identity {
   /* Stores DDO object on IPFS
    * @param {object} ddo - DDO object
    * @returns {string} an IPFS hash under which DDO is stored
-   * */
+  **/
 
+  // TODO IPFS connector should be configured only once
```
katszwn: 👍

Issue #62, opened 2018-05-29
Functionality for consuming the credentialRequest


Exulansis:
### Description

The library should expose a simple api for parsing and consuming the received credentialRequest client side.

### TODO
- [x] Define and design the desired functionality
- [x] Implement the methods

PR #64, opened 2018-05-29
Feature/#62 consuming cr

Exulansis:
Addresses #62 
Adds a method on the CR instance to check if a credential is constraint compliant.

Further changes will need to be made to the `greater` and `smaller` constraint functions to compare dates correctly.


ts/credentialRequest/index.ts
```diff
@@ -33,14 +35,25 @@ export class CredentialRequest {
     this.requestedCredentials.push({ type, constraints: credConstraints })
   }
 
+  public doesVCredSatisfyConstraints(type: string[], credential: IVerifiableCredentialAttrs): boolean {
```
katszwn: I would make that function take an array of Verifiable Credentials and call it `applyConstraints`

PR #67, opened 2018-05-30
Added the extended type, plus corresponding context

### LINKED ISSUE #66: Modify credential types
### Description

Currently all credentials are types as `Credential`, in order to be spec compliant, and more explicit, we should use a more specific type, for instance `['Credential', 'ProofOfEmailCredential']`.

### TODO
- [x] Include these types in the credential structure
- [x] Ensure they are resolved during the normalization phase
- [x] Include the new types in the `claimsMetadata` object

Exulansis:
Fixes #66

### DIFF
--- a/ts/utils/contexts.ts
+++ b/ts/utils/contexts.ts
@@ -6,6 +6,7 @@ export const defaultContext = {
   sec: 'https://w3id.org/security#',
   dc: 'http://purl.org/dc/terms/',
   xsd: 'http://www.w3.org/2001/XMLSchema#',
+  jolo: 'https://www.identity.jolocom.com/terms/',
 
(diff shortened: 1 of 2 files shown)


ts/utils/contexts.ts
```diff
@@ -16,6 +17,8 @@ export const defaultContext = {
   signatureValue: 'sec:signatureValue',
 
   Credential: 'cred:Credential',
+  ProofOfEmailCredential: 'jolo:ProofOfEmailCredential',
+  ProofOfMobilePhoneNumberCredential: 'jolo:ProofOfMobilPhoneNumberCredential',
```
katszwn: Typo

PR #74, opened 2018-05-30
Add method to get requested credential types

Exulansis:
`.getRequestedCredentialTypes` method on the CR for easier integration.

### DIFF
--- a/ts/credentialRequest/index.ts
+++ b/ts/credentialRequest/index.ts
@@ -24,6 +24,12 @@ export class CredentialRequest {
     this.callbackURL = url
   }
 
+  public getRequestedCredentialTypes(): string[] {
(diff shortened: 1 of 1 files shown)


ts/credentialRequest/index.ts
```diff
@@ -24,6 +24,12 @@ export class CredentialRequest {
     this.callbackURL = url
   }
 
+  public getRequestedCredentialTypes(): string[] {
+    const credentials = []
+    this.requestedCredentials.forEach((credential) => credentials.push(...credential.type))
+    return Array.from(new Set(credentials))
```
katszwn: Perhaps that should return an array of arrays, which would be W3C compatible credential types

Issue #63, opened 2018-05-29
Redefine exposed api


Exulansis:
### Description

The methods for creating and interacting with identities / credentials can be presented in a better way.

### TODO
- [ ] Define the new API
- [ ] Implement

PR #78, opened 2018-06-04
Class and method for assembling credential response

### LINKED ISSUE #77: Method for sending verifiable credentials to the service as part of the SSO process
### Description

Currently the service can easily specify and encode their requirements using the `credentialRequest` class. An equivalent class is required to enable the client to package, encode, and send the requested claims to the service.

### TODO
- [ ] Implement the class and the respective methods.

Exulansis:
Fixes #77 

MVP state, further polishing will be done in #63

### DIFF
--- a/ts/credentialResponse/index.ts
+++ b/ts/credentialResponse/index.ts
@@ -1,0 +1,70 @@
+import { plainToClass, classToPlain } from 'class-transformer'
+import { TokenSigner, decodeToken } from 'jsontokens'
+import { IVerifiableCredentialAttrs } from '../credentials/verifiableCredential/types'
+import { CredentialRequest } from '../credentialRequest'
+import { ISuppliedCredentialsAttrs, ICredentialResponseAttrs } from './types'
+import { areCredTypesEqual } from '../utils/general'
+
+export class CredentialResponse {
+  private iss: string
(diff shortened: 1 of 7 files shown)


ts/credentialResponse/index.ts
```diff
@@ -0,0 +1,70 @@
+import { plainToClass, classToPlain } from 'class-transformer'
+import { TokenSigner, decodeToken } from 'jsontokens'
+import { IVerifiableCredentialAttrs } from '../credentials/verifiableCredential/types'
+import { CredentialRequest } from '../credentialRequest'
+import { ISuppliedCredentialsAttrs, ICredentialResponseAttrs } from './types'
+import { areCredTypesEqual } from '../utils/general'
+
+export class CredentialResponse {
+  private iss: string
```
katszwn: I would name it `issuer`

ts/utils/general.ts
```diff
@@ -0,0 +1,3 @@
+export const areCredTypesEqual = (first: string[], second: string[]): boolean => {
```
katszwn: Maybe we could name this file `utils/index.ts`? Then it would be easier to import it and `general` is not really useful

ts/credentialResponse/index.ts
Exulansis: :+1:

PR #98, opened 2018-06-20
Signed cred requests and responses

Exulansis:
Implementation of credentialRequest / signedCredentialRequest, credentialResponse / signedCredentialResponse, including initial tests.


Exulansis:
The duplicate code issues are a bit tricky to fix, we might have to use a generic class for JWT wrapped elements, and then extend it, or use mixins...

katszwn:
I think for the duplication we might as well just put this method in a jwt util file, which I guess should be easy?

Exulansis:
Fixed the merge conflict, and some issues related to tests.

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "test": "./node_modules/.bin/nyc -e 'ts' -r lcov -r text yarn run unitTests",
     "prepublish": "rm -rf ./js/; ./node_modules/.bin/tsc -p ./tsconfig.json",
-    "unitTests": "mocha tests/*.test.ts -r ts-node/register",
+    "unitTests": "mocha tests/**/*.test.ts -r ts-node/register",
```
katszwn: 🎉

ts/credentialRequest/credentialRequest.ts
```diff
@@ -1,24 +1,36 @@
...
 
 export class CredentialRequest {
-  private requesterIdentity: string
   private callbackURL: string
   private requestedCredentials: ICredentialRequest[] = []
 
-  public setRequesterIdentity(requesterIdentity: string) {
-    this.requesterIdentity = requesterIdentity
+  public static create(args: ICredentialRequestCreationArgs): CredentialRequest {
+    const cr = new CredentialRequest()
```
katszwn: I would use here the full name - as it is now, it requires more effort for the reader to understand what the shortcut means

ts/credentialRequest/credentialRequest.ts
```diff
@@ -1,24 +1,36 @@
...
+    cr.setCallbackURL(args.callbackURL)
+    args.requestedCredentials.forEach((req) => cr.addRequestedClaim(req))
+    return cr
+  }
+
+  private addRequestedClaim({constraints, type}: {type: string[], constraints: IConstraint[]}) {
```
katszwn: Should we perhaps call it `addConstraints`?

ts/credentialRequest/credentialRequest.ts
```diff
@@ -1,24 +1,36 @@
...
+  IConstraint,
+  ICredentialRequestCreationArgs
 } from './types'
-import { areCredTypesEqual } from '../utils/credentials'
-import { ISignedCredentialAttrs } from '../credentials/signedCredential/types';
 
 export class CredentialRequest {
-  private requesterIdentity: string
   private callbackURL: string
   private requestedCredentials: ICredentialRequest[] = []
```
katszwn: It's not really an array of credentials, but of credential requests - maybe we could name it a `credentialRequest` ?

ts/credentialRequest/signedCredentialRequest/types.ts
```diff
@@ -0,0 +1,29 @@
+import { CredentialRequest } from '../credentialRequest';
+import { ICredentialRequestAttrs } from '../types';
+
+export interface IJWTHeader {
```
katszwn: Could we use full words everywhere in this file? It's not straightforward what it is :)

ts/credentialRequest/types.ts
```diff
@@ -1,5 +1,5 @@
 export interface IConstraint {
-  [operator: string]: [{ var: string }, string | comparable] | boolean[]
+  [operator: string]: boolean[] | Array<{ var: string } | string | comparable >
```
katszwn: It would be nice to extract it as a separate type

ts/credentialRequest/types.ts
```diff
@@ -21,7 +21,11 @@ export interface IExposedConstraintFunctions {
...
 export interface ICredentialRequestAttrs {
-  requesterIdentity: string
-  callbackURI: string
+  callbackURL: string
   requestedCredentials: ICredentialRequest[]
 }
+
+export interface ICredentialRequestCreationArgs {
+  callbackURL: string
+  requestedCredentials: Array<{type: string[], constraints: IConstraint[]}>
```
katszwn: That could also be a separate type

ts/credentialResponse/types.ts
```diff
@@ -1,7 +1,7 @@
 import { ISignedCredentialAttrs } from '../credentials/signedCredential/types'
 
+// TODO needed? Given it proxies
```
katszwn: So it looks like this interface is used in both request and response - it should be then reused, and name should be generic, perhaps just requestedCredentials (since even in response we use requested credentials)? We can't use `ICredentialAttrs`, right?

ts/credentialResponse/signedCredentialResponse/signedCredentialResponse.ts
```diff
@@ -0,0 +1,107 @@
...
+import { privateKeyToDID, encodeAsJWT, computeJWTSignature } from '../../utils/crypto'
+import { CredentialResponse } from '../credentialResponse'
+
+export class SignedCredentialResponse {
+  private header: IJWTHeader = {
+    alg: 'ES256K',
+    typ: 'JWT'
+  }
+
+  private payload: ISignedCredResponsePayload = {
```
katszwn: Why do we need defaults here?

ts/wallet/index.ts
```diff
@@ -35,20 +31,20 @@ export class IdentityWallet {
...
-  public createCredentialResponse(credentials: ISignedCredentialAttrs[]): string {
-    const res = new CredentialResponse().create(credentials)
-    res.setIssuer(this.did)
-    return res.toJWT(this.privateKey.privateKey)
-  }
+  // public createCredentialResponse(credentials: ISignedCredentialAttrs[]): string {
+  //   const res = new CredentialResponse().create(credentials)
+  //   res.setIssuer(this.did)
+  //   return res.toJWT(this.privateKey.privateKey)
+  // }
```
katszwn: Can't we remove those?

ts/utils/crypto.ts
```diff
@@ -29,9 +34,39 @@ export function privateKeyToDID(privateKey: Buffer): string {
...
+type jwtPayload = ISignedCredRequestPayload | ISignedCredResponsePayload
+export function encodeAsJWT(header: IJWTHeader, payload: jwtPayload, signature: string): string {
+  const jwtParts = []
+  jwtParts.push(base64url.encode(JSON.stringify(header)))
+  jwtParts.push(base64url.encode(JSON.stringify(payload)))
+  jwtParts.push(signature)
+  return jwtParts.join('.')
+}
+
+export function computeJWTSignature(payload: jwtPayload, privateKey: Buffer): string {
```
katszwn: 👍

katszwn:
Also - perhaps you could open and issue to refactor JWT relates code duplications? Should be easy to extract those to a util file.

ts/credentialRequest/credentialRequest.ts
Exulansis: That would be `CredentialRequestCreationArgs` ?

ts/credentialRequest/credentialRequest.ts
Exulansis: This adds a new claim to the request, for instance adds a `ProofOfEmailCredential`, including the constraints. So it's not only the constraints.
We could perhaps rename it to something that communicates it better.

ts/credentialRequest/credentialRequest.ts
Exulansis: It is an array of credential requests, so we could rename it to `credentialRequests` for instance.

ts/credentialRequest/signedCredentialRequest/types.ts
Exulansis: That would be `IJSONWebTokenHeader` ?

ts/credentialRequest/types.ts
Exulansis: :+1:

ts/credentialRequest/types.ts
Exulansis: :+1:

ts/wallet/index.ts
Exulansis: We should be able to.

ts/credentialRequest/credentialRequest.ts
Exulansis: Rather cr -> credentialRequest

ts/credentialRequest/signedCredentialRequest/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  header: IJWTHeader
+  payload: {
+    iat: number
+    iss: string
+    credentialRequest: ICredentialRequestAttrs
+  }
+  signature: string
+}
+
+export interface ISignedCredRequestPayload {
```
Exulansis: Remove null

PR #105, opened 2018-06-21
Feature/#83 implement identity manager

### LINKED ISSUE #83: Refactor functionality around generating HD keys to IdentityManager
### Description
Create a module of the Library that takes care of deriving HD keys according to different paths.

### TODO
- [ ] define a static `create` method which takes `seed`
- [ ] define an enum for key types
- [ ] define a member of the class - key derivation `derivationSchema` (?), which is a mapping between name and the path (`[key: string] : string`) with the basic key types
- [ ] define `getSchema()`
- [ ] define `deriveMnemonic` method, which returns mnemonic for a seed
- [ ] define `deriveKey`, which takes `name` and derives a key with a corresponding path taken from schema
- [ ] add placeholder for a method to add an entry to schema (for a custom key)

nataschaberg:
Description
Fixes #83  .

> What does it solve?

Implements class IdentityManager with according functionality.

> Were here any particular challenges that you faced and/or things that were not straightforward?

The keys schema is implemented as a member object on IdentityManager; so no auto completion is available for derivation paths.  

Checklist

- [x]  Keep flags added to the PR up to date
- [x] Make sure the title describes the code well, description should contain Fixes #number_of_an_issue and a brief decscription of changes
- [x] Make sure the PR is readable and contains changes only related to a single issue. Otherwise - divide into separate PRs
- [x] Add new tests and make sure all tests are passing
- [x] Make sure all edge cases are taken into account and covered by tests
- [ ] Run linter and fix all the errors and warnings


tests/identityManager.test.ts
```diff
@@ -0,0 +1,52 @@
...
+const expect = chai.expect
+
+describe('IdentityManager', () => {
+  const identityManager = IdentityManager.create(testKeys.default.testSeed)
+
+  describe('static create', () => {
+    it('should create an instance of IdentityManager with correct default keys schema', () => {
+      const keySchema = identityManager.getSchema()
+
+      expect(keySchema).to.have.property('jolocomIdentityKey')
```
Exulansis: Can these be omitted, and simply check for the value, e.g. if `jolocomIdentityKey` is not in the schema then

```javascript
expect(keySchema.jolocomIdentityKey).to.equal('m/73\'/0\'/0\'/0')
```
should already fail, given that the result will be `undefined`

Exulansis: Also the explicit path could perhaps be replaced with 
```javascript
keyTypes.jolocomKey
```
from the enum defined in `ts/index.ts`

tests/identityManager.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    })
+  })
+
+  describe('addSchemaEntry', () => {
+    it('should correctly add a new key schema entry to schema', () => {
+      identityManager.addSchemaEntry({name: 'gov', path: 'm/73\'/0\'/0\'/1'})
+      const keySchema = identityManager.getSchema()
+
+      expect(keySchema).to.have.property('gov')
+      expect(keySchema.gov).to.equal('m/73\'/0\'/0\'/1')
```
Exulansis: Same idea as before, is the first assertion needed?

ts/identityManager/identityManager.ts
```diff
@@ -0,0 +1,47 @@
...
+    const identityManager = new IdentityManager()
+    identityManager.seed = seed
+    identityManager.schema = {
+      ...DefaultKeyTypes
+    }
+
+    return identityManager
+  }
+
+  private deriveMnemonicFromSeed(): string {
```
Exulansis: could be an `util` function.

ts/identityManager/identityManager.ts
```diff
@@ -0,0 +1,47 @@
...
+    return keyDerivation.generateMnemonic(this.seed)
+  }
+
+  public deriveChildKeys(path: string): IKeyResponse {
```
Exulansis: Rename to `deriveChildKey`? The plural implies that it does more keys at once.

tests/identityManager.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(keySchema).to.have.property('ethereumKey')
+      expect(keySchema.jolocomIdentityKey).to.equal('m/73\'/0\'/0\'/0')
+      expect(keySchema.ethereumKey).to.equal('m/44\'/60\'/0\'/0/0')
+    })
+  })
+
+  describe('deriveChildKeys', () => {
+    it('should return a correct child key according to path', () => {
+      const keySchema = identityManager.getSchema()
+      const childKeys = identityManager.deriveChildKeys(keySchema.jolocomIdentityKey)
```
Exulansis: Can we rename this to `deriveChildKey`?

ts/identityManager/types.ts
```diff
@@ -0,0 +1,8 @@
+import { IKeyResponse } from '../utils/keyDerivation'
+import { IKeyDerivationSchema } from './identityManager';
+
+export interface IIdentityManager {
```
Exulansis: Is this necessary?

ts/identityManager/identityManager.ts
```diff
@@ -0,0 +1,47 @@
...
+export enum DefaultKeyTypes {
+  jolocomIdentityKey = 'm/73\'/0\'/0\'/0',
+  ethereumKey = 'm/44\'/60\'/0\'/0/0'
+}
+
+export interface IKeyDerivationSchema {
+  [key: string]: string
+}
+
+export class IdentityManager implements IIdentityManager {
```
Exulansis: What are the benefits of this class implementing an interface?

ts/utils/keyDerivation.ts
```diff
@@ -70,3 +71,20 @@ export function deriveEthereumKeyPair(masterKeyPair: any, path?: string): IKeyRe
...
+/* @summary - Generate a child key pair according to BIP32 specification
+ * @param {HDNode} masterKeyPair - the master keypair used to
+ *        derive the child key.
+ * @param {string} path - the path to traverse for HD derivation
+ * @returns {object} - the wif of the key and the path
+ */
+export function deriveChildKeyPair({masterKeyPair, path}: {masterKeyPair: any, path: string}): IKeyResponse {
+  const derived = masterKeyPair.derivePath(path).keyPair
+
+  return {
```
Exulansis: We are returning the public key as a buffer, but the private key as a WIF, could we return the private key as a buffer as well, so that no WIF decoding has to be done on the other end, or is there further reasoning for this?

ts/identityManager/identityManager.ts
Exulansis: Or perhaps use the function defined in utils directly.

ts/utils/keyDerivation.ts
Exulansis: We can simply add the private key as a buffer to the return values.

PR #106, opened 2018-06-21
Support for multiple claims per credential and tests

### LINKED ISSUE #104: Support multiple claims per credential
### Description

Currently each credential implicitly contains one claim, in order to enable future features, this should be changed.

### TODO
- [x] Add support for multiple claims
- [x] Add tests for the credential class.

Exulansis:
Fixes #104 
Includes tests


ts/utils/contexts.ts
```diff
@@ -27,5 +27,7 @@ export const defaultContext = {
   claim: { '@id': 'cred:claim', '@type': '@id' },
   telephone: 'schema:telephone',
   email: 'schema:email',
-  name: 'schema:name'
+  name: 'schema:name',
```
charleenfei: do we still need this 'name' if we have already family and givenName?

Issue #108, opened 2018-06-21
Implement Identity Class as member of IdentityWallet


nataschaberg:
### Description
implement get public profile on Identity class which is a member on IdentityWallet

### TODO
- [x] implement Identity class which extends DidDocument
- [x] implement identity.publicProfile.get on class
- [x] implement test for Identity class

PR #113, opened 2018-06-25
Signature validation methods

Exulansis:
Implements #108 
Signature validation methods on all verifiable primitives.

### DIFF
--- a/ts/credentialResponse/signedCredentialResponse/signedCredentialResponse.ts
+++ b/ts/credentialResponse/signedCredentialResponse/signedCredentialResponse.ts
@@ -1,5 +1,5 @@
 import 'reflect-metadata'
-import { decodeToken } from 'jsontokens'
+import { decodeToken, TokenVerifier } from 'jsontokens'
 import { classToPlain, plainToClass } from 'class-transformer'
 import { IJWTHeader } from '../../credentialRequest/signedCredentialRequest/types'
 import { ISuppliedCredentialsAttrs } from '../types'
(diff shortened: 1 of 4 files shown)


Exulansis:
We have code duplication issues again, which makes sense, this, and other similar issues can be solved by having a generic JWT wrapper class for instance.

I will try move some of the code to util files for now, to remove at least some duplication.

Exulansis:
Done.

ts/utils/jwt.ts
```diff
@@ -0,0 +1,52 @@
...
+export function encodeAsJWT(header: IJWTHeader, payload: jwtPayload, signature: string): string {
+  const jwtParts = []
+  jwtParts.push(base64url.encode(JSON.stringify(header)))
+  jwtParts.push(base64url.encode(JSON.stringify(payload)))
+  jwtParts.push(signature)
+  return jwtParts.join('.')
+}
+
+export function validateJWTSignature(jwtInstance: jwtEnabledClass, pubKey: Buffer): boolean {
+  if (!pubKey) {
```
nataschaberg: Maybe we should keep the >1 args approach consistent and use destructuring here?

ts/utils/jwt.ts
```diff
@@ -0,0 +1,52 @@
...
+  const assembledJWT = jwtInstance.toJWT()
+  return new TokenVerifier('ES256K', pubKey.toString('hex')).verify(assembledJWT)
+}
+
+// TODO Find based on key id
+export async function validateJWTSignatureWithRegistry(
+  jwtInstance: jwtEnabledClass,
+  registry: JolocomRegistry
+): Promise<boolean> {
+    if (!registry) {
```
nataschaberg: Same suggestion here with consistent approach > 1 arg.

ts/credentialResponse/signedCredentialResponse/signedCredentialResponse.ts
```diff
@@ -5,13 +5,20 @@ import { IJWTHeader } from '../../credentialRequest/signedCredentialRequest/type
 import { ISuppliedCredentialsAttrs } from '../types'
```
nataschaberg: A more general comment: should we maybe make the folder structure of credentialResponse and credentialRequest consistent with credentials? So in case of credentialRequest move credentialRequest.ts, credentialRequestParser.ts and types.ts in an own folder.

PR #142, opened 2018-07-13
Use default registry when validating

### LINKED ISSUE #141: Default Jolocom Registry when doing signature verification
### Description
Default Jolocom Registry when doing signature verification

### TODO
- [x] SignedCredential
- [x] SignedCredentialRequest
- [x] SignedCredentialResponse

katszwn:
Fixes #141 

Also, fixes a bug on Identity public key section

### DIFF
--- a/ts/utils/jwt.ts
+++ b/ts/utils/jwt.ts
@@ -45,7 +45,7 @@ export async function validateJWTSignatureWithRegistry(args: IValidateJWTSignatu
   }
 
   const issuerProfile = await registry.resolve(jwtInstance.getIssuer())
-  const pubKey = issuerProfile.getPublicKeySection()[0].publicKeyHex
+  const pubKey = issuerProfile.getPublicKeySection()[0].getPublicKeyHex()
   if (!pubKey) {
     return false
   }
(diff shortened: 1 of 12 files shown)


ts/utils/jwt.ts
```diff
@@ -45,7 +45,7 @@ export async function validateJWTSignatureWithRegistry(args: IValidateJWTSignatu
   }
 
   const issuerProfile = await registry.resolve(jwtInstance.getIssuer())
-  const pubKey = issuerProfile.getPublicKeySection()[0].publicKeyHex
+  const pubKey = issuerProfile.getPublicKeySection()[0].getPublicKeyHex()
```
katszwn: @Exulansis should take the first from public key section or match it?

ts/utils/jwt.ts
```diff
@@ -45,7 +45,7 @@ export async function validateJWTSignatureWithRegistry(args: IValidateJWTSignatu
   }
```
nataschaberg: I think it might make sense to include here the default registry as well and make the registry arg optional.

PR #164, opened 2018-09-04
Abstraction for wrapping Interaction flows in JSON Web Token

### LINKED ISSUE #162: Extract a parent class for different integration flows
### Description
The class should handle serialization to and from JSON web token format. Since all the flows will  be used as JWTs, the parent class can have a common structure of `header`, `payload` and `signature`. It should also implement common methods for signature related functionality.

The class should be extendable to enable SSO flow and receiving credentials flow.  

### TODO
- [ ] subtask

katszwn:
Fixes #162 

In order to create a new `JSONWebToken` instance, you have to call a static `create` method on the class: 
```
JSONWebToken.create({
  privateKey,
  payload: {
    iss: 'did:jolo:njkfehruu843iorj3onvgregvefd',
    iat: 3289424832,   // timestamp
    typ: InteractionType.CredentialRequest,
    credentialRequest: credentialRequestJson
  }
})
```

A `payload` needs to be of a specific type, which matches the types defined in `InteractionType`. Based on this type, a `JSONWebToken<PayloadType>` instance is created, where `PayloadType` is `SignedCredentialRequestPayload` for `CredentialRequest` type. On that instance, we're able to get an instance of the specific payload class and we get syntax completion on its methods. 

I marked old `jwt` methods as deprecated, so as not to break current implementation of `SignedCredentialRequest`. It's causing Code climate errors, but will be fixed with re-implementation of  `SignedCredentialRequest`.


ts/interactionFlows/signedCredentialRequest/signedCredentialRequestPayload.ts
```diff
@@ -0,0 +1,27 @@
+import { IPayload, InteractionType } from '../types';
+import { ISignedCredentialRequestAttrs } from '../../credentialRequest/signedCredentialRequest/types';
+import { classToPlain, plainToClass } from 'class-transformer';
+import { ISignedCredRequestPayload, ISignedCredRequestPayloadAttrs } from './types';
+import { CredentialRequest } from '../../credentialRequest/credentialRequest';
+
+export class SignedCredentialRequestPayload implements IPayload {
```
katszwn: Going to rename this class to `CredentialRequestPayload` in a follow up PR, because the signature is involved on the JWT level.

ts/interactionFlows/signedCredentialRequest/signedCredentialRequestPayload.ts
```diff
@@ -0,0 +1,27 @@
...
+  public iss: string
+  public iat: number
```
katszwn: Make both private and generate on create.

ts/utils/jwt.ts
```diff
@@ -5,29 +5,49 @@ import { ISignedCredResponsePayload } from '../credentialResponse/signedCredenti
 import { SignedCredentialRequest } from '../credentialRequest/signedCredentialRequest/signedCredentialRequest'
 import { SignedCredentialResponse } from '../credentialResponse/signedCredentialResponse/signedCredentialResponse'
 import { JolocomRegistry } from '../registries/jolocomRegistry'
+import { JSONWebToken } from '../interactionFlows/jsonWebToken';
+import { IPayload } from '../interactionFlows/types';
```
nataschaberg: just a small thing here: the import file should be JSONWebToken (JSON capital)

tests/interactionFlows/jsonWebToken.test.ts
```diff
@@ -0,0 +1,151 @@
+import { expect } from 'chai'
+import { credentialRequestJson } from '../data/credentialRequest/credentialRequest'
+import {
+  mockPrivKey,
+  mockSignedCredResponseJson,
+} from '../data/credentialResponse/signedCredentialResponse'
+import * as sinon from 'sinon'
+import * as chai from 'chai'
+import * as sinonChai from 'sinon-chai'
+import { JSONWebToken } from '../../ts/interactionFlows/jsonWebToken';
```
nataschaberg: import file: JSONWebToken (capital JSON).

ts/interactionFlows/signedCredentialRequest/signedCredentialRequestPayload.ts
katszwn: Will be addressed in a follow up PR

Issue #167, opened 2018-09-05
Re-implement CredentialRequest and CredentialResponse


katszwn:
### Description
We need to use the new JSONWebToken class. 

### TODO
- [ ] CredentialRequest
- [ ] CredentialResponse

PR #170, opened 2018-09-07
Re-implementation of credential request

katszwn:
Fixes the first part of  #167 

Now the instance of `JSONWebToken<CredentialRequestPayload>` takes the place of the `SignedCredentialRequest`. That has an implication on the API of the library: in order to create a JWT with a credential request, you ned to call `credentialRequestJSONWebToken` on the `create` of the `IdentityWallet` instance.

All the functionality previously available under `SignedCredentialRequest` specific to JWT is now under `JSONWebToken` class, and credential request specific are under `CredentialRequestPayload` class, which exposes an interface to the `CredentialRequest` class.


ts/identityWallet/identityWallet.ts
```diff
@@ -18,24 +16,20 @@ export class IdentityWallet {
...
       }
 
       return await SignedCredential.create({ credentialAttrs, privateIdentityKey: this.privateIdentityKey })
     },
-    signedCredentialRequest: (credentialRequest: CredentialRequest) =>
-      SignedCredentialRequest.create({ credentialRequest, privateKey: this.privateIdentityKey.key }),
-    signedCredentialResponse: (credentialResponse: CredentialResponse) =>
-      SignedCredentialResponse.create({ credentialResponse, privateKey: this.privateIdentityKey.key })
+    credentialRequestJSONWebToken: (payload: ICredentialRequestPayloadAttrs) => {
+      return JSONWebToken.create({privateKey: this.privateIdentityKey.key, payload})
```
katszwn: Here we should pass the private key object and use the key ID as `iss`

tests/data/interactionFlows/jsonWebToken.ts
```diff
@@ -17,19 +17,17 @@ export const credentialRequestJson = {
...
-  signature: 'SK7pVnsZkqsR_CopE35DJLvUsOgTZlFSEztNDjAAYLDjsvaoD5HrLytDWKBwr7MoUaUjEYV92Gu2br-k6lgV6Q'
+  signature: 'LyZytGL7Ixf3ulHIoHRJqQPjXqrqTKC462NRt4P6m_pyo5ROrjzrqJVSvAqhY6aHucAi2p9j16kuZKj79HUUHA'
 }
 
 export const jwtCreateArgs = {
   privateKey: Buffer.from(mockPrivKey, 'hex'),
   payload: {
-    iss: 'did:jolo:njkfehruu843iorj3onvgregvefd',
-    iat: 0,
     typ: InteractionType.CredentialRequest.toString(),
```
Exulansis: The `toString` part does not seem to be needed, removing it results in no errors and tests still pass.

tests/data/interactionFlows/jsonWebToken.ts
```diff
@@ -47,10 +45,30 @@ export const jwtCreateArgs = {
...
+b2xvOmlzc3VlciJdfV19XSwiY2FsbGJhY2tVUkwiOiJodHRwOi8vdGVzdC5jb20\
+ifSwiaWF0IjowLCJpc3MiOiJkaWQ6am9sbzo4Zjk3N2U1MGI3ZTVjYmRmZWI1M2\
+EwM2M4MTI5MTNiNzI5NzhjYTM1YzkzNTcxZjg1ZTg2Mjg2MmJhYzhjZGViIn0.L\
+yZytGL7Ixf3ulHIoHRJqQPjXqrqTKC462NRt4P6m_pyo5ROrjzrqJVSvAqhY6aH\
+ucAi2p9j16kuZKj79HUUHA'
+
+export const credentialRequestPayloadJson = {
+  iss: 'did:jolo:8f977e50b7e5cbdfeb53a03c812913b72978ca35c93571f85e862862bac8cdeb',
+  iat: 0,
+  typ: InteractionType.CredentialRequest.toString(),
```
Exulansis: Same as above

PR #189, opened 2018-09-21
Nested claims

### LINKED ISSUE #173: Support for custom nested claims.
### Description

We should allow the creation of credentials with nested claim section. The metadata construct we currently have needs to be updated as well.

### TODO
- [ ] Implement

katszwn:
Fixes #173 

Updated the version of the `cred-types-jolocom-core` library that introduces nested claims and added a test.

### DIFF
--- a/tests/data/credential/credential.ts
+++ b/tests/data/credential/credential.ts
@@ -9,6 +9,14 @@ export const multipleClaimsCreationArgs = {
   familyName: 'Rusu'
 }
 
+export const addressClaimCreationArgs = {
+  address: {
+    street: 'Kopenicker Str 147',
+    postalCode: '10829',
+    city: 'Berlin'
+  }
+}
+
 export interface BirthDateClaimInterface extends BaseMetadata {
   claimInterface?: {
     birthDate: number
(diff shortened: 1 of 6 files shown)


ts/credentials/credential/types.ts
```diff
@@ -1,16 +1,16 @@
 import { validContextEntry } from 'cred-types-jolocom-core'
 
-type validClaimTypes = string | number | boolean
-export type validClaimEntry = validClaimTypes | {[key: string]: validClaimEntry}
+type ClaimType = string | number | boolean | {}
```
charleenfei: isn't this a repeat of the type from credential-types-jolocom-core?

katszwn:
@charleenfei Thank you! Applied :)

tests/data/credential/credential.ts
```diff
@@ -17,12 +25,28 @@ export interface BirthDateClaimInterface extends BaseMetadata {
...
+  }
+}
+
 export const customClaimMetadata: BirthDateClaimInterface = {
   context: ['http://test.com', { test: 'http://test.com/terms' }],
   type: ['Credential', 'MockCredential'],
   name: 'Mock'
 }
 
+export const nestedAddressClaimMetadata: NestedAddressClaimInterface = {
```
charleenfei: maybe we should also use the default contexts here from the module in order to make the usage pattern clear?

katszwn:
Good point, will address in a follow up PR

PR #205, opened 2018-10-02
Feature/#200 auth sig verification

### LINKED ISSUE #200: Implement signature verification on authentication response
### Description
Implement the signature verification for authentication flow
### TODO
- [x] Implement logic
- [x] Fix/add tests
- [x] Add integration test for the authentication flow

nataschaberg:
Fixes #200 

Simplifies the authentication flow to use the capabilities of JWT for signing. 
Removes the need to introduce a deviation from current set up for payload needing to sign data.


tests/integration/index.ts
```diff
@@ -195,4 +198,114 @@ describe('Integration Test', () => {
...
+        }
+      })
+      authRequestJWT = authRequestJWTClass.encode()
+
+      expect(authRequestJWTClass.getPayload()).to.be.an.instanceOf(AuthenticationPayload)
+      expect(authRequestJWTClass.getPayload().authentication).to.be.an.instanceOf(Authentication)
+    })
+
+    it('Should allow for simple consumption of authentication request and generate authentication response by user', async () => {
+      sinon.stub(jr, 'createJolocomRegistry').returns(jolocomRegistry)
```
katszwn: An integration test can't stub anything - otherwise it wouldn't test the whole flow.

tests/integration/index.ts
```diff
@@ -195,4 +198,114 @@ describe('Integration Test', () => {
...
+      })
+      authResponseJWT = authResponseJWTClass.encode()
+
+      expect(authRequest).to.be.an.instanceOf(AuthenticationPayload)
+      expect(authRequest.authentication).to.be.an.instanceOf(Authentication)
+      expect(authResponseJWTClass).to.be.instanceOf(JSONWebToken)
+    })
+
+    it('Should allow for authentication response consumption & validation and credentialsReceive creation by service', async () => {
+      sinon.stub(jr, 'createJolocomRegistry').returns(jolocomRegistry)
```
katszwn: The same here

tests/integration/index.ts
```diff
@@ -195,4 +198,114 @@ describe('Integration Test', () => {
...
+      
+      credentialReceiveJWT = credReceiveJWTClass.encode()
+
+      expect(credentialFromService).to.be.an.instanceOf(SignedCredential)
+      expect(credReceiveJWTClass).to.be.an.instanceOf(JSONWebToken)
+      expect(credReceiveJWTClass.getPayload()).to.be.an.instanceOf(CredentialsReceivePayload)
+    })
+
+    it('Should allow for consumption of credentialsReceieve with correct signed credential by user', async () => {
+      sinon.stub(jr, 'createJolocomRegistry').returns(jolocomRegistry)
```
katszwn: And here

ts/interactionFlows/types.ts
```diff
@@ -37,7 +37,6 @@ export enum InteractionType {
   CredentialRequest = 'credentialRequest',
   CredentialResponse = 'credentialResponse',
   CredentialsReceiving = 'credentialsReceiving',
```
katszwn: That's not used, right? We should probably remove it :)

tests/integration/index.ts
nataschaberg: We need to 'stub' it here as we want the jolocom registry to use same ipfs and ethereum connectors as set up at beginning. Method validateSignatureWithPublicKey on JSONWebToken calls createJolocomRegistry() internally.

ts/interactionFlows/types.ts
nataschaberg: Not entirely sure what to remove here? We use the InteractionType enum.

katszwn: We have now both - `CredentialsReceiving` and `CredentialsReceive`. One needs to go away ;)

tests/integration/index.ts
katszwn: Than the interface of the verification methods is incorrect - registry should always be optionally injectable. We need to add it as an optional param in every method that uses registry under the hood.

Will open an issue for that.

katszwn: https://github.com/jolocom/jolocom-lib/issues/206

tests/integration/index.ts
katszwn: https://github.com/jolocom/jolocom-lib/issues/206

tests/integration/index.ts
katszwn: https://github.com/jolocom/jolocom-lib/issues/206

katszwn:
Since my concern will be addressed in a separate issue (https://github.com/jolocom/jolocom-lib/issues/206), the only thing is removing the obsolete InteractionType. Other than that - looks good!

PR #210, opened 2018-10-08
Change jsonld imports to use dist directory

### LINKED ISSUE #209: Import jsonld as node6 version to avoid errors while importing the library
### Description
In order to avoid errors while importing the library in the react-native environment, we should change how we import `jsonld` package - not the standard way with `lib/jsonld.js` an an entry file, but through `dist/jsonld.js` - which is webpack and node compatible. 

### TODO
- [ ] change all import statements to use `dist/jsonld.js`

katszwn:
Fixes #209


Exulansis:
Looks good!

PR #220, opened 2018-10-15
Converting date to ISO format to avoid timezone issues

Exulansis:
We need to explicitly convert dates to ISO formats in our code, otherwise things go wrong with `class-transformer` and during the canonicalization process the dates are replaced with local equivalents, which results in different times / hashes, and subsequently different digest.

Solves the issue of credentials signed in one time zone using Jolocom Lib being unverifiable in other time zones.

### DIFF
--- a/ts/credentials/signedCredential/signedCredential.ts
+++ b/ts/credentials/signedCredential/signedCredential.ts
@@ -1,5 +1,5 @@
 import 'reflect-metadata'
-import { plainToClass, classToPlain, Type, Exclude, Expose } from 'class-transformer'
+import { Transform, plainToClass, classToPlain, Type, Exclude, Expose } from 'class-transformer'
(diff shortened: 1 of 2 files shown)


nataschaberg:
👍

ts/registries/jolocomRegistry.ts
```diff
@@ -120,17 +119,19 @@ export class JolocomRegistry {
...
+
     try {
       const identity = await this.resolve(did)
-      pubKey = identity.getPublicKeySection()
-        .find(pubKeySection => pubKeySection.getIdentifier() === keyId)
+      pubKey = identity.getPublicKeySection().find(pubKeySection => pubKeySection.getIdentifier() === keyId)
+
+      if (!pubKey) {
+        return false
+      }
```
katszwn: 👍

Issue #230, opened 2018-11-05
Implement replay attack protection for JWT


nataschaberg:
### Description
Implement replay attack protection for JWT

### TODO
- [x] implement logic
- [x] fix/add tests

PR #231, opened 2018-11-06
Feature/#230 replay attack protection jwt

nataschaberg:
Implemenation of #230


tests/identityWallet/identityWallet.test.ts
```diff
@@ -61,11 +57,12 @@ describe('IdentityWallet', () => {
...
     const sandbox = sinon.createSandbox()
     let stubCredCreate
-    let stubFromJWTEncodable
+    let spyFromJWTEncodable
+    let interactionToken
 
     before(() => {
       stubCredCreate = sandbox.stub(SignedCredential, 'create').callsFake(() => stubbedCredential)
-      stubFromJWTEncodable = sandbox.stub(JSONWebToken, 'fromJWTEncodable').callsFake(stubbedFromJWTEncodable)
+      spyFromJWTEncodable = sandbox.spy(JSONWebToken, 'fromJWTEncodable')
```
Exulansis: +1

Exulansis: +1

tests/interactionTokens/jsonWebToken.test.ts
```diff
@@ -66,23 +71,31 @@ describe('JSONWebToken', () => {
...
     const jwt = new JSONWebToken()
-
+    const nonce = Math.random().toString(36)
+    
     jwt.setSignature(signature)
     jwt.setTokenContent(credReq)
     jwt.setTokenIssuer(iss)
+    jwt.setTokenTimeStamps()
+    jwt.setTokenNonce(nonce)
+    jwt.setTokenAudience('did:jolo:testauddid')
```
Exulansis: Can we replace this with a did from the *.data.ts files?
Something like mockDID

ts/interactionTokens/JSONWebToken.ts
```diff
@@ -117,7 +152,14 @@ export class JSONWebToken<T extends JWTEncodable> implements IDigestable {
   */
 
   public static decode<T extends JWTEncodable>(jwt: string): JSONWebToken<T> {
-    return JSONWebToken.fromJSON(decodeToken(jwt))
+    const jwtClass = JSONWebToken.fromJSON(decodeToken(jwt))
+    const validTimePeriod = jwtClass.getExpirationTime() - jwtClass.getIssueTime() > 1
```
Exulansis: Can we instead do something like
```typescript
if (jwtClass.getExpirationDate() < Date.now()) {
  ...
}
```

I will add the change and check if it breaks tests.

tests/interactionTokens/jsonWebToken.test.ts
nataschaberg: 👍

ts/interactionTokens/JSONWebToken.ts
nataschaberg: good catch

Issue #270, opened 2019-02-06
Credential constraint issues


chunningham:
### Description
1. It is possible to assemble a credential response in such a way that when it's sent to the service, it will cause the `satisfiesRequests` to throw an error (which is not handled by the lib). The error will bubble into "userland", and if the developer does not handle it, lead to a crash.

2. It is possible to bypass all credential requirements if no credentials are sent. If an empty array is set, the `satisfiesRequest` function will always return true, no matter what the `credentialConstraints` are.

### TODO
- [x] write tests for these cases
- [x] fix these cases

PR #271, opened 2019-02-06
Bug/#270 credential constraint fixes

chunningham:
fixes for #270


Exulansis:
That looks better, thank you!

PR #274, opened 2019-02-11
Update MINOR and PATCH versions of dependencies. That should be backw…

palinka:
…ards-compatible following semver.

yarn test runs fine.


palinka:
I used `yarn upgrade-interactive --latest` to do the updates.

palinka:
Reorganizing dependency PR. No merging & closing.

Tasks will be addressed in: #277

PR #275, opened 2019-02-11
buildSystem/breaking semver dependency upgrades

palinka:
This upgrades major and 0.x.y releases for our dependencies which do NOT break tests. The branch is called "breaking..", because major releases MAY break things. 

Please check if all is still good.


palinka:
Reorganizing dependency PR. No merging & closing. 

Tasks will be addressed in: https://github.com/jolocom/jolocom-lib/issues/277

PR #287, opened 2019-04-02
286/migrate to eslint add prettier

saifahn:
Adds ESLint, removes TSLint file.

There are a number of new dependencies:

- @typescript-eslint/eslint-plugin allows typescript specific rules to be used in eslint
- @typescript-eslint/parser allows eslint to parse typescript correctly
- eslint-config-prettier turns off rules that would conflict with prettier
- eslint-plugin-prettier allows eslint and prettier to be run in one command -> prettier is run as part of eslint

Each of these have their own dependencies (1-4 of them). Is this acceptable?

Additionally, I added comments in the `.eslintrc.js` as notes as to what each rule does, should these remain?


saifahn:
The latest commit is a big one as it formats all the code according to prettier.

PR #289, opened 2019-04-02
Bump TS version, and type the parse object to ensure resulting declaration file is correct.

### LINKED ISSUE #288: Incorrect declaration file generated.
**Describe the bug**
`yarn prepublish` produces a faulty `index.d.ts` declaration file.

**To Reproduce**
1. Clone the repository
2. Run `yarn install; yarn prepublish`
3. Open the `./js/index.d.ts` file using an IDE and you should see the incorrect typing on the `parse` object.

**Expected behavior**
No error.

Exulansis:
Explicitly typed the problematic section and bumped the typescript version to 3.4.1

Fixes #288

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "sinon": "^7.2.5",
     "source-map-support": "^0.5.10",
     "ts-node": "^8.0.2",
-    "typescript": "^3.3.3",
+    "typescript": "^3.4.1",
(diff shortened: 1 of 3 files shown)


saifahn:
I've taken a look at the .js file before and after this change and the errors are no longer present after this fix.

PR #276, opened 2019-02-12
[WIP] Experimental/payments

nataschaberg:
Implements #265 #267 #273 
Implementation of payment interaction flow

PR #292, opened 2019-04-08
264/revisit payments tests

Exulansis:
Cleaned up implementation for #276.


Exulansis:
After this #287 can be updated and merged.

Issue #300, opened 2019-04-16
Seed Phrase Recovery


VolkerSchiewe:
**Is your feature request related to a problem? Please describe.**
Key recovery is a problem that every key management system should face. Seed phrase (mnemonic) recovery is a very simple method that many cryptowallets are using.

**Describe the solution you'd like**
The jolocom-lib should provide a method to recover the identity wallet based on the mnemonic phrase. Additionally the mnemonic phrase can be integrated in the creation of of the identity.

**Additional context**
Mnemonics are based on [bip39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)

chunningham:
merged, closing 👍

PR #304, opened 2019-04-17
Seed phrase recovery

VolkerSchiewe:
I think it is a good idea to move every thing that is mnemonic related to the lib, but I'm not sure if adding the `getMnemonic` method to the key provider is a good idea as it is kind of similar to the `getPrivateKey`-antipattern, isn't it?

cloeses #300


VolkerSchiewe:
@chunningham I moved the recovery method to be part of the vkp and works basically like a constructor for it.

PR #240, opened 2018-11-08
Feature/#238 use accessors

Exulansis:
Results in clearer documentation, potentially simpler API in the future.

PR #310, opened 2019-05-23
Update DID Document implementation

VolkerSchiewe:


mnzaki:
We might still want to use `AuthenticationSection` I think, in some cases we might want to inline the publicKey instead of just refer to it

mnzaki:
I'm basic the DDO migration work on this branch

ts/identity/didDocument/didDocument.ts
```diff
@@ -258,7 +324,7 @@ export class DidDocument implements IDigestable {
    * @example `const didDocument = DidDocument.fromPublicKey(Buffer.from('abc...ffe', 'hex'))`
    */
 
-  public static fromPublicKey(publicKey: Buffer): DidDocument {
+  public static async fromPublicKey(publicKey: Buffer): Promise<DidDocument> {
```
chunningham: this doesnt need to be async does it? none of the calls inside it are

ts/identity/didDocument/types.ts
```diff
@@ -1,17 +1,21 @@
...
   IAuthenticationSectionAttrs,
+  IPublicKeySectionAttrs,
   IServiceEndpointSectionAttrs,
 } from './sections/types'
 import { ILinkedDataSignatureAttrs } from '../../linkedDataSignature/types'
 import { ContextEntry } from 'cred-types-jolocom-core'
 
 export interface IDidDocumentAttrs {
-  '@context': ContextEntry[]
+  '@context': ContextEntry[] | string
```
chunningham: does this need to be `'@context'` or could it just be `context` and have the classtoplain handle it?

VolkerSchiewe: These are the type definitions for the json-ld DID Document, so I think it has to be `@context`

tests/data/didDocument.data.ts
```diff
@@ -24,23 +25,79 @@ export const didDocumentJSON = {
...
 }
 
 export const normalizedDidDocument =
-  '\
-<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> <https://w3id.org/security#owner> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> .\n\
-<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <http://purl.org/dc/terms/created> "1970-01-01T00:00:00.000Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .\n\
-<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <https://w3id.org/security#authenticationMethod> _:c14n0 .\n\
-<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <https://w3id.org/security#publicKey> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> .\n\
-_:c14n0 <https://w3id.org/security#publicKey> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> .\n'
+  '<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#jolocomPubProfile> <http://schema.org/description> "Verifiable Credential describing entity profile" .\n' +
```
Exulansis: Some statements are missing here
- Missing `type` for publicProfile / service endpoint
- First authentication entry (key #1) is a string literal and not a reference to the section (lack of `<>`). Not sure if an issue
- `type` predicate missing in auth key sections and public key sections

ts/identity/didDocument/didDocument.ts
```diff
@@ -17,37 +18,67 @@ import {
...
+   * NOTE: the context from jolocom identities is automatically replaced with
+   * the latest from the library when deserializing from JSON
    * @see {@link https://json-ld.org/spec/latest/json-ld/#the-context | JSON-LD context}
    * @example `console.log(didDocument.context) // [{name: 'http://schema.org/name', ...}, {...}]`
    */
 
   @Expose({ name: '@context' })
-  get context(): ContextEntry[] {
-    return this['_@context']
+  @Transform(
```
Exulansis: Is this needed? The context is already being replaced during signature validation / generation, is there another case where the context matters?

VolkerSchiewe: I guess it is needed. In case of migration we will parse the old DID Document and transform it with the class transformer into the new format. This should also add the new context for the new format. The updated DID Document can than be signed and pushed to the registry.
I think its not a good to change the context during validation as this changes the way  the JSON-LD is normalized and therefore the signature.

ts/identity/didDocument/didDocument.ts
```diff
@@ -84,9 +115,12 @@ export class DidDocument implements IDigestable {
    * @example `console.log(didDocument.authentication) // [AuthenticationSection {...}, ...]`
    */
 
-  @Expose()
-  @Type(() => AuthenticationSection)
```
Exulansis: Because this was lost, the normalization is not correct. If an instance of a `PublicKey` section is passed, it's not passed through class transformer, and the plain JSON is populated.

tests/data/didDocument.data.ts
Exulansis: Solved in #340

ts/identity/didDocument/didDocument.ts
Exulansis: Addressed in #340

ts/identity/didDocument/didDocument.ts
Exulansis: Consolidated in #240

Exulansis:
#340 Addresses most of the comments raised here.

Issue #309, opened 2019-05-23
Support for render metadata in credential offer


Exulansis:
## Description
In order to allow for better rendering / UX of credential offers on client devices (e.g. mobile wallets), issuers must be able to provide metadata for styling purposes (e.g. logo).

Updating the `requestedInput` field is **not in scope** for this issue.

## To be done:
- [x] Define data structure for credential offers ([current status](https://gist.github.com/Exulansis/e422f97ec8c32c70bd7bc6677cf29e93))
- [x] Define data structure for credential offer responses
- [x] Update the interaction tokens according to the new data structure (`requestedInput` acts as a placeholder for now)
- [x] Adjust tests
- [x] Document the new flows

_Note: The current data structure for the `renderInfo` field is meant to be compatible with the [credential manifest](https://identity.foundation/schemas/claims/ClaimManifest) as defined by the DIF. At later points in time, the `@context` field can be introduced to normalize the key names_

Exulansis:
Last thing left for now is documentation.

Exulansis:
Documentation has been added, [section 5](https://jolocom-lib.readthedocs.io/en/latest/interactionFlows.html) updated.

PR #311, opened 2019-05-23
309/credential offer update

Exulansis:
Implements #309


tests/integration/credentialOffer.integration.ts
```diff
@@ -140,7 +143,7 @@ describe('Integration Test - Token interaction flow Credential Offer', () => {
         jolocomRegistry,
       )
     } catch (err) {
-      expect(true).to.be.false
+      return expect(true).to.be.false
```
mnzaki: You can get better errors with something like this:
```
it('Should allow for consumtion of valid credential receive token by user', done => {
/* ... */
userIdentityWallet.validateJWT(
        decodedCredReceive,
        credOfferResponseJWT,
        jolocomRegistry,
      ).then(done, done)
/* ... */
})

saifahn:
Waiting for functionality on SmartWallet before updating both.

tests/integration/credentialOffer.integration.ts
Exulansis: :+1: Thank you, was looking for something like this.

PR #314, opened 2019-06-12
docs: add note about usage in browsers and react native

mnzaki:


mnzaki:
@chunningham that note actually needs updating by now :smile:

PR #320, opened 2019-07-10
New DID Registry Integration

VolkerSchiewe:
Main changes:
* I did not saw the reasoning for the Identity Class in the old implementation as it was just holding the DID Document and the Public profile. Now It holds all the data that belongs to the identity (kind of an abstraction of DID Document). The Identity class can be converted to a DID Document. We could also include the logic of the identity class in the DID Document and remove the identity.
* The public key is stored on chain now and used to validate identity updates. Therefore the ethereum key and the identity key need to be the same. Hence I switched it everywhere.

Open Todos:
- [ ] add authorization section to DID Document
- [ ] Fix payment integration test
- [x] Migration Strategy


VolkerSchiewe:
Will not be integrated like this. closing for now.

Issue #322, opened 2019-07-25
Modify key provider encryption method to use IVs


Exulansis:
Currently methods associated with encryption / decryption on the Key Provider use the `aes256-cbc` cipher with no initialization vectors.

Given that the current method is marked as deprecated (in node.js at least), and that the wallet uses IVs internally, we should adapt these methods to ensure we have a consistent implementation.

### TODO:
- [x] Update the methods (i.e. [here](https://github.com/jolocom/jolocom-lib/blob/develop/ts/vaultedKeyProvider/softwareProvider.ts#L168))
- [x] Update unit tests

PR #323, opened 2019-07-25
Updated encrypt / decrypt implementations to use IVs

Exulansis:
Updated the SoftwareKeyProvider class and it's associated tests.

Implements #322

### DIFF
--- a/docs/gettingStarted.rst
+++ b/docs/gettingStarted.rst
@@ -62,18 +62,17 @@ How to create a self-sovereign identity
 
 In broad strokes, the creation of a self-sovereign identity comprises the following steps:
 
-* Instantiate the ``SoftwareKeyProvider`` class by providing a 32 byte random seed ``Buffer``, and a password for encryption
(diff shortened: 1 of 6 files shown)


docs/gettingStarted.rst
```diff
@@ -62,18 +62,17 @@ How to create a self-sovereign identity
...
 
-The following sections present these steps in greater detail.
+The following sections elaborate on these steps.
 
 **Instantiate the Key Provider class**
 
 The ``SoftwareKeyProvider`` class abstracts all functionality related to `deriving key pairs <https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki>`_ and creating / validating cryptographic signatures.
-The 32 byte seed used to instantiate the key provider is persisted in the instance, encrypted using the provided password.  Therefore, all operations which involve key derivation require the password as well.
-This is what instantiating a key provider looks like:
+Currently two ways of instantiating the class are supported, namely using the constructor or using the static ``fromSeed`` method:
```
VolkerSchiewe: Just as a side note: also recover key pair creates a new softwareKeyProvider and takes the mnemonic as argument. But it's probably more confusing than helpful at this point.

Issue #324, opened 2019-07-29
Remove `generateRandomId` function from crypto utils


Exulansis:
### Description

It is now superseded by the `getRandomBytes` function exposed by the key provider.

PR #325, opened 2019-07-29
Removed duplicate function generating random bytes, updated tests

Exulansis:
Implements #324

### DIFF
--- a/tests/identityWallet/validateJWT.test.ts
+++ b/tests/identityWallet/validateJWT.test.ts
@@ -9,11 +9,8 @@ import { JSONWebToken } from '../../ts/interactionTokens/JSONWebToken'
 import { DidDocument } from '../../ts/identity/didDocument/didDocument'
 import {
   invalidSignature,
-  validSignature,
   validSignedCredReqJWT,
-  validSignedCredResJWT,
-  invalidNonce,
-  validNonce,
+  invalidNonce, validSignedCredResJWT
 } from '../data/interactionTokens/jsonWebToken.data'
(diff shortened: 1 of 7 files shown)


ts/utils/crypto.ts
```diff
@@ -26,25 +26,3 @@ export function publicKeyToDID(publicKey: Buffer): string {
...
 }
-
-/**
- * Generates a "random" sequence of bytes
- * @param nrOfBytes - The size of the output
- * @ignore
- */
-
-/* TODO Revert to previous implementation once compatibility issues are sorted */
-export function generateRandomID(nrOfBytes: number): string {
```
VolkerSchiewe: We could deprecated this method instead of removing it.Then this wouldn't be a breaking change. But if we publish it together with the IV-KeyProvider change it makes no difference.

tests/identityWallet/validateJWT.test.ts
```diff
@@ -49,38 +47,71 @@ describe('IdentityWallet validate JWT', () => {
...
+      payload: {
+        ...validSignedCredResJWT.payload,
+        nonce: invalidNonce,
+      },
+    }
+
+    /** @dev Restored in afterEach */
+    sandbox.stub(SoftwareKeyProvider, 'verifyDigestable').resolves(true)
+
+    console.log(tokenWIthInvalidNonce)
```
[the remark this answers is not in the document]
Exulansis: Thanks, removed.

ts/utils/crypto.ts
Exulansis: It's only used internally.

PR #327, opened 2019-07-31
Add support for resolving identities based on DID method prefix

### LINKED ISSUE #326: Enable support for resolving using multiple DID methods.
## Description

Currently the library is designed with one DID method in mind, i.e. we intend one `registry` to be instantiated, and used throughout the relying applications. Furthermore, in some cases (e.g. the `publicKeyToDID` function, the `DidDocument.fromPublicKey` function, the `did:jolo` DID prefix is assumed), which can cause further interoperability issues.

## TODO

- [ ] Provide a generalized `registry` interface that can be extended with extra registries, and used to resolve DIDs "dynamically" based on the prefix.
- [ ] Add unit tests.

Exulansis:
Resolves #326


tests/integration/payment.integration.ts
```diff
@@ -116,8 +116,7 @@ describe('Integration Test - EXPERIMENTAL Token interaction flow Payment', () =>
         jolocomRegistry,
       )
     } catch (err) {
-      console.log('ERROR: ', err)
-      expect(true).to.be.false
+      expect(true).to.eq(true)
```
mnzaki: To get better error logging you can use something like this: 
```
  it('.....', done => {
    /* .... */
   serviceIdentityWallet.validateJWT(
        decodedPaymentResponse,
        paymentRequestJWT,
        jolocomRegistry,
      ).then(() => done()).catch(done)
  })
```

Or if `validateJWT` doesn't return anything it could have also been `.then(done, done)`

mnzaki:
1. Are we ever going to need multiple *registries*? Multiple instances? Different classes altogether? How would that interact (ex. does it make it redundant?) with having multiple resolvers in a registry?

2. Why is the `MultiResolver` interface coupled with the `JolocomRegistry` in a way? I mean it is not used in the `IRegistry` interface, and the `JolocomRegistry` constructor explicitly expects a `MultiResolver` (as opposed to an abstract `IResolver` or something). I'm not entirely sure, but I think every registry (or at least a majority) is going to need access to a resolver, in which case it would make sense to add that to `IRegistry.create`.

ts/registries/jolocomRegistry.ts
```diff
@@ -259,7 +265,16 @@ export const createJolocomRegistry = (
...
+
+  const validatingJolocomResolver = createValidatingResolver(
+    createJolocomResolver(ethereumConnector, ipfsConnector),
+  )(noValidation)
+
+  const multiResolver = new MultiResolver({
+    jolo: validatingJolocomResolver,
+  })
+
+  const jolocomRegistry = new JolocomRegistry(multiResolver)
```
mnzaki: The resolver is not configurable through the `createJolocomRegistry` interface, unlike all the other configurable bits. Also it is assigned through the constructor, unlike the others below. I think it is better to use the constructor actually, but perhaps keep things consistent now and change it all together later.

ts/resolver/index.ts
```diff
@@ -0,0 +1,103 @@
...
+import { getMethodPrefixFromDid } from '../utils/crypto'
+import { IEthereumConnector } from '../ethereum/types'
+import { IIpfsConnector } from '../ipfs/types'
+
+/**
+ * Curried function for assembling a resolver to be used as a {@link ValidatingDidResolver} in the {@link MultiResolver}.
+ *   Curried to take a {@link DidDocumentResolver}, followed by a {@link DidDocumentValidator}
+ */
+
+export const createValidatingResolver: ValidatingDidResolver = resolver => (
```
mnzaki: Is the currying useful here or are you currying by default now :D

mnzaki:
For reference, this addresses #326

mnzaki:
If the client application is interacting with different DID methods then they probably will need to use the library specific to each each. Should `jolocom-lib` really be responsible for supporting other methods? It makes sense to be able to resolve other methods as well though, just so we can do things with their documents, for "interop", but it feels like an external responsibility.

With this in mind, perhaps a much simpler interface can be sufficient?
Simply passing a `resolver: (did: string) => IDidDocumentAttrs | null` argument to the registry, which if provided will be tried first to resolve everything (regardless of method). If it returns null, the registry itself tries to resolve the DID (using just its own method).
This would allow for other uses (besides multi-method support) as well, like for example resolving some special DIDs locally or through a cache and falling back to the registry otherwise.

tests/integration/payment.integration.ts
mnzaki: Actually is this on purpose? I just noticed that you also changed `false` to `true`

mnzaki:
More thoughts on this: the registry itself never actually needs to resolve a non-jolo DID, as far as I could tell. It's the `identityWallet` that needs to potentially resolve other methods some times. Perhaps the mult-did-resolve concern should be in the identityWallet?

ts/resolver/index.ts
Exulansis: Good point, this doesn't need to be curried.

Exulansis:
> More thoughts on this: the registry itself never actually needs to resolve a non-jolo DID, as far as I could tell. It's the identityWallet that needs to potentially resolve other methods some times. Perhaps the mult-did-resolve concern should be in the identityWallet?

Agree, as we are introducing the resolver, it becomes apparent that the Registry should not need to handle any resolution (besides when making sure that the identity being created exists). There's a comment on the `resolve` method of the registry that this API call will be deprecated soon in favor of a resolver.

For now I've tried to keep the public API the same, to introduce no breaking changes at early stages. But we should decide on how the multi resolver is exposed to the user.

tests/integration/payment.integration.ts
Exulansis: Not on purpose, it's a mistake, thanks for pointing it out.

It should be changed to the way you suggested, long term we need some consistent way to test async stuff. Chai-as-promised could be something to look into.

mnzaki:
From the session we had today:
- `MultiResolver` configurable through `createJolocomWallet` and not coupled to registry
  - registry takes a resolution function
- functions taking registry actually need resolver
    - `identityWallet.validateJWT`
    - `utils/validation.ts`
      - `validateDigestable/s`

tests/integration/payment.integration.ts
Exulansis: Addressed, thank you.

ts/identityWallet/identityWallet.ts
```diff
@@ -478,18 +478,22 @@ export class IdentityWallet {
...
    * @param sendJWT - optional send JSONWebToken Class which is used to validate the token nonce and the aud field on received token
-   * @param customRegistry - optional custom registry
+   * @param resolver - optional custom resolver used to fetch signer public keys
    */
 
   public async validateJWT<T extends JWTEncodable, A extends JWTEncodable>(
     receivedJWT: JSONWebToken<T>,
     sendJWT?: JSONWebToken<A>,
-    customRegistry?: IRegistry,
+    resolver: MultiResolver = mutliResolver,
```
mnzaki: Same thoughts about using `MultiResolver` interface instead of a simpler `resolve(did: string): Promise<IDidDocumentAttrs>`: it's probably unnecessarily complicated, and potentially limiting.

As we'd discussed earlier, the methods that need a registry to just resolve things should simply take a resolver function.

The `MultiResolver` itself may be useful, but I don't think we should constrain the `IdentityWallet` (or other classes) to need it specifically. So maybe use it by passing just its `resolve` method?

Exulansis: We might need different resolution functions depending on the JWT. All supported resolution methods are amassed in the multi-resolver. If this is to take only a resolution function, we would have to always select the right resolution function before making this call.

I am currently looking at changing the interface of the `MultiResolver` to be:
`resolve(did: string): Promise<Identity>`
This will allow it to be integrated much easier across the wallet / lib.

This would probably add an extra function, e.g. `parser` that the user can define when creating their resolution function. The `parser` will expect the same args returned by the `resolutionFn` they pass initially. So they can choose how to fetch public profiles and other stuff we can't account for in the `resolution` function, validate them using the `validator`, and use their `parser`(name should be probably changed), to, once validated, get the output of the `resolver` and return and `Identity`

mnzaki: > We might need different resolution functions depending on the JWT. All supported resolution methods are amassed in the multi-resolver. If this is to take only a resolution function, we would have to always select the right resolution function before making this call.

I meant for `resolve(did: string): Promise<IDidDocumentAttrs>` to resolve all DID methods, in the same way as `MultiResolver.resolve`, not that it is a resolution function of one specific method. My point is about the interface complexity, I don't (yet?) see the advantage of enforcing an interface with `supportedMethods`, `resolutionMap`, and `resolve` methods. In the use cases I see currently it seems all that needs to be done is just call the `resolve` method, and it will either succeed or not.

Again let me reword: I'm talking about not explicitly requiring a `MultiResolver` instance as an argument, because that doesn't seem to be useful (and enforces an unnecessarily more complex interface). `MultiResolver` itself may be useful as a tool, and maybe it should stick around, but the methods that currently take an entire registry as a parameter can be simplified to simply take a resolution function

Exulansis: > My point is about the interface complexity, I don't (yet?) see the advantage of enforcing an interface with supportedMethods, resolutionMap, and resolve methods. In the use cases I see currently it seems all that needs to be done is just call the resolve method, and it will either succeed or not.

I see what you mean now. That makes sense.

mnzaki:
Testing out the interface diff script on this:
```diff
diff -U 3 -r develop/js/identity/didDocument/didDocument.d.ts 274ac95/js/identity/didDocument/didDocument.d.ts
--- develop/js/identity/didDocument/didDocument.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/identity/didDocument/didDocument.d.ts	2019-08-29 14:30:14.000000000 +0200
@@ -31,7 +31,7 @@
     addPublicKeySection(section: PublicKeySection): void;
     addServiceEndpoint(endpoint: ServiceEndpointsSection): void;
     resetServiceEndpoints(): void;
-    static fromPublicKey(publicKey: Buffer): Promise<DidDocument>;
+    static fromPublicKey(publicKey: Buffer, didBuilder?: import("../../utils/crypto").DidBuilder): DidDocument;
     sign(vaultedKeyProvider: IVaultedKeyProvider, derivationArgs: IKeyDerivationArgs, keyId: string): Promise<void>;
     digest(): Promise<Buffer>;
     normalize(): Promise<string>;
diff -U 3 -r develop/js/identityWallet/identityWallet.d.ts 274ac95/js/identityWallet/identityWallet.d.ts
--- develop/js/identityWallet/identityWallet.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/identityWallet/identityWallet.d.ts	2019-08-29 14:30:13.000000000 +0200
@@ -8,8 +8,9 @@
 import { CredentialRequest } from '../interactionTokens/credentialRequest';
 import { IKeyMetadata, ISignedCredCreationArgs } from '../credentials/signedCredential/types';
 import { ITransactionEncodable } from '../contracts/types';
-import { IRegistry } from '../registries/types';
 import { CredentialOfferRequest } from '../interactionTokens/credentialOfferRequest';
+import { MultiResolver } from '../resolver';
+import { DidDocument } from '../identity/didDocument/didDocument';
 export declare class IdentityWallet {
     private _identity;
     private _publicKeyMetadata;
@@ -18,7 +19,7 @@
     private _contractsGateway;
     did: string;
     identity: Identity;
-    didDocument: import("../identity/didDocument/didDocument").DidDocument;
+    didDocument: DidDocument;
     publicKeyMetadata: IKeyMetadata;
     private vaultedKeyProvider;
     constructor({ identity, publicKeyMetadata, vaultedKeyProvider, contractsGateway, contractsAdapter, }: IIdentityWalletCreateArgs);
@@ -33,7 +34,7 @@
     private createPaymentReq;
     private createPaymentResp;
     private initializeAndSign;
-    validateJWT<T extends JWTEncodable, A extends JWTEncodable>(receivedJWT: JSONWebToken<T>, sendJWT?: JSONWebToken<A>, customRegistry?: IRegistry): Promise<void>;
+    validateJWT<T extends JWTEncodable, A extends JWTEncodable>(receivedJWT: JSONWebToken<T>, sendJWT?: JSONWebToken<A>, resolver?: MultiResolver): Promise<void>;
     private sendTransaction;
     transactions: {
         sendTransaction: (request: ITransactionEncodable, pass: string) => Promise<string>;
diff -U 3 -r develop/js/index.d.ts 274ac95/js/index.d.ts
--- develop/js/index.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/index.d.ts	2019-08-29 14:30:14.000000000 +0200
@@ -14,8 +14,8 @@
         constraintFunctions: import("./interactionTokens/interactionTokens.types").IExposedConstraintFunctions;
         fuelKeyWithEther: typeof fuelKeyWithEther;
         getIssuerPublicKey: typeof getIssuerPublicKey;
-        validateDigestable: (toValidate: import("./linkedDataSignature/types").IDigestable, customRegistry?: import("./registries/types").IRegistry) => Promise<boolean>;
-        validateDigestables: (toValidate: import("./linkedDataSignature/types").IDigestable[], customRegistry?: import("./registries/types").IRegistry) => Promise<boolean[]>;
+        validateDigestable: (toValidate: import("./linkedDataSignature/types").IDigestable, resolver?: import("./resolver").MultiResolver) => Promise<boolean>;
+        validateDigestables: (toValidate: import("./linkedDataSignature/types").IDigestable[], resolver?: import("./resolver").MultiResolver) => Promise<boolean[]>;
     };
     KeyTypes: typeof KeyTypes;
 };
diff -U 3 -r develop/js/registries/jolocomRegistry.d.ts 274ac95/js/registries/jolocomRegistry.d.ts
--- develop/js/registries/jolocomRegistry.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/registries/jolocomRegistry.d.ts	2019-08-29 14:30:14.000000000 +0200
@@ -1,21 +1,23 @@
 import { IIpfsConnector } from '../ipfs/types';
 import { IEthereumConnector } from '../ethereum/types';
 import { IdentityWallet } from '../identityWallet/identityWallet';
-import { SignedCredential } from '../credentials/signedCredential/signedCredential';
 import { Identity } from '../identity/identity';
 import { IRegistryCommitArgs, IRegistryStaticCreationArgs, IRegistry } from './types';
 import { IVaultedKeyProvider, IKeyDerivationArgs } from '../vaultedKeyProvider/types';
 import { IContractsAdapter, IContractsGateway } from '../contracts/types';
+import { ValidatingIdentityResolver } from '../resolver/types';
 export declare class JolocomRegistry implements IRegistry {
     ipfsConnector: IIpfsConnector;
     ethereumConnector: IEthereumConnector;
     contractsAdapter: IContractsAdapter;
     contractsGateway: IContractsGateway;
+    readonly resolver: ValidatingIdentityResolver;
+    private readonly didBuilder;
+    constructor(resolver: ValidatingIdentityResolver, didBuilder: any);
     create(vaultedKeyProvider: IVaultedKeyProvider, decryptionPassword: string): Promise<IdentityWallet>;
     commit(commitArgs: IRegistryCommitArgs): Promise<void>;
     resolve(did: any): Promise<Identity>;
     authenticate(vaultedKeyProvider: IVaultedKeyProvider, derivationArgs: IKeyDerivationArgs, did?: string): Promise<IdentityWallet>;
-    fetchPublicProfile(entry: string): Promise<SignedCredential>;
     private resolveSafe;
 }
 export declare const createJolocomRegistry: (configuration?: IRegistryStaticCreationArgs) => JolocomRegistry;
diff -U 3 -r develop/js/registries/types.d.ts 274ac95/js/registries/types.d.ts
--- develop/js/registries/types.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/registries/types.d.ts	2019-08-29 14:30:13.000000000 +0200
@@ -4,6 +4,8 @@
 import { IVaultedKeyProvider, IKeyDerivationArgs } from '../vaultedKeyProvider/types';
 import { Identity } from '../identity/identity';
 import { IContractsAdapter, IContractsGateway } from '../contracts/types';
+import { ValidatingIdentityResolver } from '../resolver/types';
+import { DidBuilder } from '../utils/crypto';
 export interface IRegistryStaticCreationArgs {
     contracts: {
         adapter: IContractsAdapter;
@@ -11,6 +13,8 @@
     };
     ipfsConnector: IIpfsConnector;
     ethereumConnector: IEthereumConnector;
+    didResolver?: ValidatingIdentityResolver;
+    didBuilder?: DidBuilder;
 }
 export interface IRegistryCommitArgs {
     vaultedKeyProvider: IVaultedKeyProvider;
Only in 274ac95/js: resolver
diff -U 3 -r develop/js/utils/crypto.d.ts 274ac95/js/utils/crypto.d.ts
--- develop/js/utils/crypto.d.ts	2019-08-29 13:55:51.000000000 +0200
+++ 274ac95/js/utils/crypto.d.ts	2019-08-29 14:30:13.000000000 +0200
@@ -1,3 +1,8 @@
 /// <reference types="node" />
 export declare function sha256(data: Buffer): Buffer;
-export declare function publicKeyToDID(publicKey: Buffer): string;
+declare type DigestFunction = (toDigest: Buffer) => Buffer;
+export declare type DidBuilder = (publicKey: Buffer) => string;
+export declare const publicKeyToDID: (prefix: string) => (digestFunction: DigestFunction) => DidBuilder;
+export declare const publicKeyToJoloDID: DidBuilder;
+export declare const getMethodPrefixFromDid: (did: string) => string;
+export {};
diff -U 3 -r develop/js/utils/validation.d.ts 274ac95/js/utils/validation.d.ts
--- develop/js/utils/validation.d.ts	2019-08-29 13:55:52.000000000 +0200
+++ 274ac95/js/utils/validation.d.ts	2019-08-29 14:30:13.000000000 +0200
@@ -1,4 +1,5 @@
 import { IDigestable } from '../linkedDataSignature/types';
-import { IRegistry } from '../registries/types';
-export declare const validateDigestable: (toValidate: IDigestable, customRegistry?: IRegistry) => Promise<boolean>;
-export declare const validateDigestables: (toValidate: IDigestable[], customRegistry?: IRegistry) => Promise<boolean[]>;
+import { MultiResolver } from '../resolver';
+export declare const validateDigestable: (toValidate: IDigestable, resolver?: MultiResolver) => Promise<boolean>;
+export declare const validateDigestables: (toValidate: IDigestable[], resolver?: MultiResolver) => Promise<boolean[]>;
+export declare const noValidation: <T>(toValidate: T) => Promise<boolean>;
```

ts/registries/jolocomRegistry.ts
```diff
@@ -196,7 +196,7 @@ export class JolocomRegistry implements IRegistry {
     did?: string,
   ): Promise<IdentityWallet> {
     const publicIdentityKey = vaultedKeyProvider.getPublicKey(derivationArgs)
-    did = did ? did : publicKeyToDID(publicIdentityKey)
+    did = did ? did : this.didBuilder(publicIdentityKey)
```
mnzaki: Again with the rotated public key issue. Maybe the `did` argument should be required? Maybe the key provider should store the original (before rotation) entropy as well (so we can always derive the DID)?
Again not sure what to do exactly, but this line just seems like an invitation for application layer bugs later on (simply not caring about passing a DID, and then major issues later)

Issue #329, opened 2019-08-02
Validate signed JSON documents


VolkerSchiewe:
**Description**
As the DID Document class for example should be able to parse different versions of DID Documents it will not be possible to use it for validation anymore. It should be possible to validate signed JSON objects without parsing them into 

related to #308

PR #335, opened 2019-08-07
Enable signature validation on arbitrary JsonLD documents

Exulansis:
Implements #329


ts/validation/jsonLdValidator.ts
```diff
@@ -0,0 +1,80 @@
...
+  public normalize = normalizeJsonLD
+}
+
+export const normalizeJsonLD = async (
+  data: JsonLdObject,
+  context = defaultContext,
+) => {
+
+  return canonize(data, {
+    expandContext: context
```
VolkerSchiewe: Why are you expanding the context here? I would expect that this can crash the validation of a DID Document e.g. as the signature was created based on the original context.

ts/validation/validation.ts
```diff
@@ -0,0 +1,73 @@
...
+  toValidate: IDigestible[],
+  customRegistry?: IRegistry,
+): Promise<boolean[]> =>
+  Promise.all(
+    toValidate.map(async digestable =>
+      validateDigestible(digestable, customRegistry),
+    ),
+  )
+
+export const validateJsonLd = async (
```
VolkerSchiewe: As this method is nearly a copy of `validateDigestible` it can just convert the `json`object to `JsonLdDigestible` and call `validateDigestible` therewith.

Exulansis: Good point, addressed.

ts/validation/jsonLdValidator.ts
Exulansis: This is addressed in the [context dedicated PR](https://github.com/jolocom/jolocom-lib/pull/338/files#diff-0e2b062e2de7d6dbf5393ca5f0469166L72).

Exulansis: Merging the context PR into this branch now, to make reviewing easier. Will mark this as resolved.

Exulansis:
Closing in favor of a cleaner #340

PR #340, opened 2019-08-11
Support generic jsonld signature validation

Exulansis:
Implements #329


ts/vaultedKeyProvider/softwareProvider.ts
```diff
@@ -200,30 +200,30 @@ export class SoftwareKeyProvider implements IVaultedKeyProvider {
...
    * Digest the passed object, and validate the signature using a provided public key
-   * @param toVerify - Instance of class that implements IDigestable
+   * @param toVerify - Instance of class that implements IDigestible
    * @param publicKey - Public key used to generate the signature
-   * @example `await SoftwareKeyProvider.verifyDigestable(publicKey, publicProfileSignedCredential) // true`
+   * @example `await SoftwareKeyProvider.verifyDigestible(publicKey, publicProfileSignedCredential) // true`
    */
 
-  public static async verifyDigestable(
+  public static async verifyDigestible(
```
mnzaki: breaking change?

ts/identity/didDocument/didDocument.ts
```diff
@@ -408,12 +400,16 @@ export class DidDocument implements IDigestable {
    */
 
   public static fromJSON(json: IDidDocumentAttrs): DidDocument {
-    const options: ClassTransformOptions | undefined = json.id.startsWith(
-      'did:jolo',
-    )
-      ? { version: json.specVersion || 0 }
-      : undefined
+    return plainToClass(DidDocument, json, {
+      version: getDidDocVersion(json),
```
VolkerSchiewe: The implementation returned undefined ClassTransformerOption if the DID was not jolo. Now we are passing always `version: 0` in case it is not a jolo DID. I would assume that this can lead to problems when we try to parse a DID Document of an other DID Method which has already the latest spec format.

ts/identity/didDocument/types.ts
```diff
@@ -5,17 +5,18 @@ import {
   IServiceEndpointSectionAttrs,
 } from './sections/types'
 import { ILinkedDataSignatureAttrs } from '../../linkedDataSignature/types'
-import { ContextEntry } from 'cred-types-jolocom-core'
+import { SignedJsonLdObject } from '../../validation/jsonLdValidator'
 
-export interface IDidDocumentAttrs {
-  '@context': ContextEntry[] | string
```
VolkerSchiewe: Why has the context been removed?

ts/linkedDataSignature/suites/ecdsaKoblitzSignature2016.ts
```diff
@@ -138,21 +138,14 @@ export class EcdsaLinkedDataSignature
...
-
-    json['@context'] = defaultContext
-
-    delete json.signatureValue
-    delete json.type
-    delete json.id
-
-    return canonize(json)
+  public async normalize() {
+    return normalizeLdProof(this.toJSON(), this._context)
```
VolkerSchiewe: As soon as we add migrations to that class(e.g. because of spec changes), adding a default context will break the signature validation

ts/utils/contexts/contextTransformers.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * {@link ContextTransformer} that attempts to replace any URLs in the context array
+ *  with entries from a local cache.
+ *  @note Does not recurse!
+ *  @param context - The context to be transformed
+ *  @returns - Context with urls replaced with cached vocabularies, or left unmodified
+ *    if no local record is found
+ */
+
+export const cachedContextTransformer: ContextTransformer = context => {
```
VolkerSchiewe: adding types to the `context` parameter would make it easier to understand the line below

ts/utils/contexts/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const credentialCustomTerms = {
+  cred: 'https://w3id.org/credentials#',
+  claim: { '@id': 'cred:claim', '@type': '@id' },
+  Credential: 'cred:Credential',
+  issuer: { '@id': 'cred:issuer', '@type': '@id' },
+  issued: { '@id': 'cred:issued', '@type': 'xsd:dateTime' },
+}
+
+interface ContextMap {
```
VolkerSchiewe: I'm not fully understanding what the purpose of these 3 fields should be. May a ts-docs comment could help?

ts/utils/contexts/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  extendedDefinitions: [commonCustomTerms, identityCustomTerms],
+}
+
+export const didDocumentContext = identityContextMap.specMandated
+  .concat(identityContextMap.published)
+  .concat(identityContextMap.extendedDefinitions)
+
+export const credentialContextMap: ContextMap = {
+  specMandated: [],
+  published: ['https://w3id.org/did/v0.11', 'https://w3id.org/did/v1'],
```
VolkerSchiewe: Why does the creadential context contain the urls of the DID Document contexts? shouldn't it be `https://www.w3.org/2018/credentials/v1`?

ts/utils/contexts/localCache.ts
```diff
@@ -1,48 +1,21 @@
...
-    expires: { '@id': 'sec:expiration', '@type': 'xsd:dateTime' },
-    proof: { '@id': 'sec:proof', '@type': '@id' },
-    EcdsaKoblitzSignature2016: 'sec:EcdsaKoblitzSignature2016',
-    created: { '@id': 'dc:created', '@type': 'xsd:dateTime' },
-    creator: { '@id': 'dc:creator', '@type': '@id' },
-    domain: 'sec:domain',
-    nonce: 'sec:nonce',
-    signatureValue: 'sec:signatureValue',
+export const locallyCachedContexts: { [k: string]: JsonLdContextEntry } = {
+  'https://www.w3.org/2019/did/v1': {
```
VolkerSchiewe: We should probably throw an error in case someone has this in his DID Document :grin:

ts/identity/didDocument/types.ts
```diff
@@ -5,17 +5,18 @@ import {
...
   id: string
-  authentication?:
-    | IAuthenticationSectionAttrsv0[]
-    | IAuthenticationSectionAttrs[]
+  authentication?: DidDocAuthenticationSection
   publicKey?: IPublicKeySectionAttrs[]
   service?: IServiceEndpointSectionAttrs[]
   created?: string
-  proof?: ILinkedDataSignatureAttrs
+  proof: ILinkedDataSignatureAttrs
```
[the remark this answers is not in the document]
chunningham: It should be, but making it optional gets us in a real pickle regarding the `DidDocument.digest()` method and with ClassTransformer as well. Some serious reshuffling of all the related types could be involved

ts/identity/didDocument/types.ts
chunningham: replaced by the context field brought in by extending `SignedJsonLdObject`

ts/utils/contexts/contextTransformers.ts
chunningham: I think the declaration for `arrayContext` is fairly explanatory

ts/identity/didDocument/types.ts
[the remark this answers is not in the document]
mnzaki: @Exulansis said the spherity (so uport?) documents don't even have a proof section anyway, and maybe `.digest` should fail if there's no proof section?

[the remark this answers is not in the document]
chunningham: (correct me if I'm wrong) in the case of `fromJSON` on a DDO without a proof, one will be generated, and if we `toJSON` that DDO then it will include this random proof which wont add up. As long as we don't `toJSON` then we wont notice. I wouldn't call this a show stopper for this PR but it is something we should note

PR #342, opened 2019-08-14
Downgrade bip39 dependency

VolkerSchiewe:
Because react native uses an old version of the JS Core it does not support some modern JS features. An example for that is `String.prototype.normalize()` which is used in `bip39.js` starting from version 3.0.0. In earlier versions of this library they use the package `unorm` which provides that feature also for earlier versions of node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "dependencies": {
     "base64url": "^3.0.1",
     "bip32": "^1.0.2",
-    "bip39": "^3.0.1",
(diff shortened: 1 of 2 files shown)


mnzaki:
why not add a shim for `String.prototype.normalize` instead of downgrading though?

VolkerSchiewe:
I tried to polyfill this yesterday but it did not worked out, but now I found a way which works fine :+1:

PR #359, opened 2019-09-07
Moved validation logic to validateJWT. Slight refactor

### LINKED ISSUE #358: Moved all validation logic for JSONWebTokens to validateJWT
### Description

Currently, the validation of a token's expiry date happens when it is decoded / used to instantiate a `JSONWebToken` class. Other validation logic (i.e. signature, nonce, aud) happens in `identityWallet.validateJWT`.

We can move the check for the expiry date to the aforementioned function as well, this way expired tokens can still be decoded for debugging purposes, and all validation logic is delegated to the `validateJWT` function.

Exulansis:
Closes #358.

### DIFF
--- a/ts/interactionTokens/JSONWebToken.ts
+++ b/ts/interactionTokens/JSONWebToken.ts
@@ -15,7 +15,7 @@ import { CredentialResponse } from './credentialResponse'
 import { CredentialRequest } from './credentialRequest'
 import { Authentication } from './authentication'
 import { CredentialsReceive } from './credentialsReceive'
-import { handleValidationStatus, keyIdToDid } from '../utils/helper'
+import { keyIdToDid } from '../utils/helper'
 import { PaymentResponse } from './paymentResponse'
 import { PaymentRequest } from './paymentRequest'
 import { CredentialOfferResponse } from './credentialOfferResponse'
@@ -194,10 +194,7 @@ export class JSONWebToken<T extends JWTEncodable> implements IDigestable {
    */
 
   public static decode<T extends JWTEncodable>(jwt: string): JSONWebToken<T> {
-    const interactionToken = JSONWebToken.fromJSON(decodeToken(jwt))
-    handleValidationStatus(interactionToken.expires > Date.now(), 'exp')
-
-    return interactionToken as JSONWebToken<T>
+    return JSONWebToken.fromJSON<T>(decodeToken(jwt))
   }
 
   /*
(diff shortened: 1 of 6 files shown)


ts/interactionTokens/JSONWebToken.ts
```diff
@@ -194,10 +194,7 @@ export class JSONWebToken<T extends JWTEncodable> implements IDigestable {
    */
 
   public static decode<T extends JWTEncodable>(jwt: string): JSONWebToken<T> {
```
mnzaki: Now the decode function does not check for expiry any more? Is this intended? Is this breaking? Do we check for expiry somewhere else?

Exulansis: If you look at the diff/ changes, you can see that the validation code for the expiry has been moved to where other validation happens (`validateJWT`). The tests were updated as well. I don't see how this would break any existing code. Decoding a token could throw an error before, because of the expiration, now it can't. So if any error handling was in place there, it won't fire. But no new errors or unexpected flows are introduced.

If they ever validate the JWT (by calling the `validateJWT` method), then they have to account for false / true (now it has one more reason to return false.).

Exulansis: I've left a longer comment in the issue itself. Decode now only decodes, and `validateJWT` does the validation.

Exulansis:
> I think the logic enclosed in `handleValidationState` is nice to have contained in something reusable. It can possibly go where all the validation logic will go (or just kept around now till moved).

I think the `handleValidationState` is not too reusable. It is defined as
```
(condition, key) => !condition 
    ? throw new Error(errorMap[key])
    : undefined
```

So it takes a bool, and throws an error in case the bool is false, with an error message from a local map (which is currently tailored for token validation).

Exulansis:
HandleValidationState currently abstracts a `if` statement + encapsulates a map of error messages.

ts/interactionTokens/JSONWebToken.ts
mnzaki: It does make sense to separate decoding and validation, I was just curious about whether this would break anything (in case something else depends on it). Can't think of anything particular myself anyway

mnzaki:
> HandleValidationState currently abstracts an `if` statement + encapsulates a map of error messages.

I think we do need to encapsulate the validation errors though, because external code will want to find out what kind of error it is (so probably some check with error codes from an enum). Although the current implementation of `handleValidationState` wasn't really useful to that end, especially because the error codes are not exported and are flattened out to the error messages as well... so yeah, off with its head

Exulansis:
> > HandleValidationState currently abstracts an `if` statement + encapsulates a map of error messages.
> 
> I think we do need to encapsulate the validation errors though, because external code will want to find out what kind of error it is (so probably some check with error codes from an enum). Although the current implementation of `handleValidationState` wasn't really useful to that end, especially because the error codes are not exported and are flattened out to the error messages as well... so yeah, off with its head

+1 I think an issue can be created to have some better overarching way to do this. Error code + reusing messages where suitable is a good idea.

PR #361, opened 2019-09-18
Commit Process: husky + precise-commits (prettier) + commitizen

### LINKED ISSUE #360: Improve commit process
### Description
Use `commitizen` and `prettier` through `husky`

Related to #348

mnzaki:
Closes #360


mnzaki:
Note: in a future update we should probably run eslint as well, but first we should actually clean up the code :eyes:

PR #371, opened 2019-09-25
369/custom expiry date

### LINKED ISSUE #369: Support user defined expiration dates
### Description
Add support for passing custom expiry dates when creating *Signed Credentials* and *Interaction Tokens*.

Currently the lib fills in the values automatically (to 1 year for Credentials, and 1 hour for Interaction tokens).

### TODO
- [x] Allow library users to pass the expiry date as part of Credential / Token creation flow.
- [x] Add unit tests to ensure the values are set correctly.
- [ ] Document the new API.

Exulansis:
closes #369

### DIFF
--- a/ts/credentials/signedCredential/signedCredential.ts
+++ b/ts/credentials/signedCredential/signedCredential.ts
@@ -321,8 +321,11 @@ export class SignedCredential implements IDigestable {
      */
     signedCredential.expires =
       expires || new Date(new Date().setFullYear(new Date().getFullYear() + 1))
-
     signedCredential.issued = new Date()
+
+    if (signedCredential.expires <= signedCredential.issued) {
+      throw new Error('Expiry date should be greater than current date')
+    }
     signedCredential.prepareSignature(issInfo.keyId)
     signedCredential.issuer = issInfo.issuerDid
 
--- a/ts/credentials/signedCredential/signedCredential.ts
+++ b/ts/credentials/signedCredential/signedCredential.ts
@@ -21,6 +21,10 @@ import { ISigner } from '../../registries/types'
 import { Credential } from '../credential/credential'
 import { SoftwareKeyProvider } from '../../vaultedKeyProvider/softwareProvider'
 
+
+// Credentials are valid for a year by default
+const DEFAULT_EXPIRY_MS = 365 * 24 * 3600 * 1000
+
 /**
  * @description Data needed to prepare signature on credential
  * @see {@link https://w3c.github.io/vc-data-model/ | specification}
@@ -319,8 +323,7 @@ export class SignedCredential implements IDigestable {
      * @dev setFullYear returns an unix timestamp, not sure if there's a less
      * convoluted way to get the current date + 1 year
      */
-    signedCredential.expires =
-      expires || new Date(new Date().setFullYear(new Date().getFullYear() + 1))
+    signedCredential.expires = expires || new Date(Date.now() + DEFAULT_EXPIRY_MS)
     signedCredential.issued = new Date()
 
     if (signedCredential.expires <= signedCredential.issued) {
--- a/ts/credentials/signedCredential/signedCredential.ts
+++ b/ts/credentials/signedCredential/signedCredential.ts
@@ -1,20 +1,20 @@
 import 'reflect-metadata'
 import {
-  plainToClass,
   classToPlain,
-  Type,
-  Expose,
   Exclude,
+  Expose,
+  plainToClass,
   Transform,
+  Type,
 } from 'class-transformer'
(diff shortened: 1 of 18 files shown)


ts/interactionTokens/JSONWebToken.ts
```diff
@@ -177,14 +180,23 @@ export class JSONWebToken<T extends JWTEncodable> implements IDigestable {
...
+    expiry: Date = new Date(Date.now() + EXPIRY_MS),
+  ) {
+    const issued = new Date()
+
+    if (expiry <= issued) {
+      throw new Error('Expiry date should be greater than current date')
+    }
+
+    this.payload.iat = issued.getTime()
+    this.payload.exp = expiry.getTime() || this.payload.iat + EXPIRY_MS
```
mnzaki: Something is off here, it will always use `expiry.getTime()` (or crash if expiry is not `Date`-like). I think also it would be helpful if this utility would rather take an expiry length and add it to the issued time, that way it's actually useful more than just two setters in one :sweat_smile:

ts/interactionTokens/JSONWebToken.ts
```diff
@@ -177,14 +180,23 @@ export class JSONWebToken<T extends JWTEncodable> implements IDigestable {
...
+  /**
    * @description - Populates the token issued and exiry times, expiry defaults to 1 hr
+   * @param expiry - Expiration date - {@link Date} instance
    * @returns {void}
    */
 
-  public setIssueAndExpiryTime() {
-    this.payload.iat = Date.now()
-    this.payload.exp = this.payload.iat + 3600000
+  public timestampAndSetExpiry(
```
mnzaki: breaking change, please bring back old interface and deprecate it.
but why the name change? Also this just feels like 2 setters in 1, maybe `expires` and `issued` should also get normal setters

ts/credentials/signedCredential/signedCredential.ts
```diff
@@ -301,22 +301,27 @@ export class SignedCredential implements IDigestable {
...
     const signedCredential = SignedCredential.fromJSON(json)
 
-    const expiry = new Date()
-    expiry.setFullYear(expiry.getFullYear() + 1)
+    /**
+     * @dev setFullYear returns an unix timestamp, not sure if there's a less
+     * convoluted way to get the current date + 1 year
+     */
+    signedCredential.expires =
+      expires || new Date(new Date().setFullYear(new Date().getFullYear() + 1))
```
mnzaki: Why not just? `expires + 365 * 24 * 3600 * 1000`
I don't think expiry should be in terms of the calendar year anyway

ts/interactionTokens/JSONWebToken.ts
mnzaki: I pushed a commit to bring back `setIssueAndExpiryTime` https://github.com/jolocom/jolocom-lib/pull/371/commits/df8111c6a5c2d98e6f463072477b0d3dda0756df

Exulansis: Thank you, I will add a deprecation message to the comments.
I will also mark the method as `internal`, since it's not really inteneded for library end users, and is only public so it can be accessed by the `identityWallet` when creating signed credentials.

ts/credentials/signedCredential/signedCredential.ts
Exulansis: In this case if the user specifies a custom expiry date, it will also be incremented by one year. This assignment could be split over two lines, will try it out and see if that's better.

mnzaki: eh no that's a brainfart, I meant to say "Why not just `expires || Date.now() + 365 * 24 * 3600 * 1000`"

ts/interactionTokens/JSONWebToken.ts
Exulansis: I thought about doing this as well. What I could not decide on is how the interval / length should be represented, i.e. seconds, milliseconds.

ts/credentials/signedCredential/signedCredential.ts
Exulansis: Addressed in f666d3b

Exulansis: The expiry on signed credentials is expected to be a Date instance, so the expression above had to be wrapped by the new Date constructor.

ts/interactionTokens/JSONWebToken.ts
Exulansis: Good catch, looks like I was trying to do 2 things at once there. I refactored the code slightly to be less convoluted.

PR #376, opened 2019-11-22
Feat/329/jsonld validation

chunningham:
key files are `ts/linkedData/index.ts` and `ts/linkedData/types.ts`. Other changes are including jsonLD types in DID doc sections and credential types, and using the digestion/validation functions in DID doc and signed credential classes.

### DIFF
--- a/tests/data/didDocument.data.ts
+++ b/tests/data/didDocument.data.ts
@@ -40,6 +40,7 @@ export const didDocumentJSON = {
 export const normalizedDidDocument =
   '\
 <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> <https://w3id.org/security#owner> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> .\n\
+<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> <https://w3id.org/security#publicKeyHex> "03848af62bffceb57631780ac0e0726106ee1c23262d6fd7ef906559d68f53a551" .\n\
 <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <http://purl.org/dc/terms/created> "1970-01-01T00:00:00.000Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .\n\
 <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <https://w3id.org/security#authenticationMethod> _:c14n0 .\n\
 <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> <https://w3id.org/security#publicKey> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> .\n\
--- a/ts/linkedData/index.ts
+++ b/ts/linkedData/index.ts
@@ -1,18 +1,11 @@
 import { SoftwareKeyProvider } from '../vaultedKeyProvider/softwareProvider'
 import { ILinkedDataSignatureAttrs } from '../linkedDataSignature/types'
-import {
-  getIssuerPublicKey,
-  keyIdToDid
-} from '../utils/helper'
+import { getIssuerPublicKey, keyIdToDid } from '../utils/helper'
 import { IRegistry } from '../registries/types'
-import { registries } from '../registries/index'
+import { registries } from '../registries/'
 import { sha256 } from '../utils/crypto'
 import { canonize } from 'jsonld'
-import {
-  JsonLdObject,
-  SignedJsonLdObject,
-  JsonLdContext,
-} from './types'
+import { JsonLdObject, SignedJsonLdObject, JsonLdContext } from './types'
 
 /**
  * Helper function to handle JsonLD normalization.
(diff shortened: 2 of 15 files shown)


ts/linkedData/index.ts
```diff
@@ -0,0 +1,89 @@
...
+ *  the '@context' on the data will be discarded.
+ * @param data - {@link JsonLdObject} without the '@context' section
+ * @param context - JsonLD context to use during normalization
+ */
+
+export const normalizeJsonLd = async (
+  { ['@context']: _, ...data }: JsonLdObject,
+  context: JsonLdContext,
+) => canonize(data, {
+  expandContext: context,
```
VolkerSchiewe: replacing the context can always result in breaking the signature validation. This is why I'm afraid of this context expansion.

ts/linkedData/types.ts
```diff
@@ -0,0 +1,16 @@
+import { ILinkedDataSignatureAttrs } from '../linkedDataSignature/types'
+
+type JsonLdPrimitive = string | number | boolean | JsonLdObject | JsonLdObject[]
+
+export type JsonLdContextEntry = string | JsonLdObject
+
+export type JsonLdContext = JsonLdContextEntry | JsonLdContextEntry[]
+
+export interface JsonLdObject {
+  '@context'?: JsonLdContext
```
VolkerSchiewe: The context is the key element of a JsonLD, isn't it? Hence I do not see why this is an optional parameter

chunningham: contexts are the killer app, but they're not actually mandated by the spec. any valid JSON object is also a valid JSON-LD object, context or not

ts/linkedData/index.ts
chunningham: context is provided as a second param so that contexts can be used for objects which may not have their own, e.g. normalizing a json-ld proof. An overload with just one arg which doesn't do this replacement may be good though

VolkerSchiewe: Yes, replacing the context might be a nice feature for now (because may contexts are just garbage), but for long term we should avoid this... I would vote for a second optional parameter to override the context and use the existing one otherwise.

ts/linkedData/index.ts
```diff
@@ -0,0 +1,89 @@
...
+      json.proof.creator,
+      issuerIdentity.didDocument,
+    )
+    return SoftwareKeyProvider.verify(
+      await digestJsonLd(json, json['@context']),
+      issuerPublicKey,
+      Buffer.from(json.proof.signatureValue, 'hex'),
+    )
+  } catch {
+    return false
```
Exulansis: False can mean either failed validation, or failed resolution (i.e. no internet). Should this function throw in case resolution fails?

*Note: validateDigestable attempts to resolve the signer outside of a try / catch block.*

ts/linkedData/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Helper function to handle JsonLD normalization.
+ * @dev The function expects the JsonLD '@context' to be passed as an argument,
+ *  the '@context' on the data will be discarded.
+ * @param data - {@link JsonLdObject} without the '@context' section
+ * @param context - JsonLD context to use during normalization
+ */
+
+export const normalizeJsonLd = async (
```
Exulansis: Mark as @internal? Also applies to other functions exported for internal usage but not meant to be used directly.

ts/linkedData/index.ts
Exulansis: Also does this need unit tests?

ts/utils/contexts.ts
```diff
@@ -1,8 +1,8 @@
-import { ContextEntry } from 'cred-types-jolocom-core'
+import { JsonLdContextEntry } from '../linkedData/types'
 
 /* Expanded default context for verifiable credentials, more verbose, but works in offline use cases */
 
-export const defaultContext: ContextEntry[] = [
+export const defaultContext: JsonLdContextEntry[] = [
```
Exulansis: The type can be inferred by TS here, which allows us to not export the `JsonLdContextEntry` type at all, and only use the more general `JsonLdContext` type.

tests/data/didDocument.data.ts
```diff
@@ -40,6 +40,7 @@ export const didDocumentJSON = {
 export const normalizedDidDocument =
```
Exulansis: Some things are missing in the normalized output, i.e. the types of the keys in the `publicKey` and `authentication` sections.

tests/data/didDocument.data.ts
```diff
@@ -40,6 +40,7 @@ export const didDocumentJSON = {
...
   '\
 <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> <https://w3id.org/security#owner> <did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777> .\n\
+<did:jolo:b2d5d8d6cc140033419b54a237a5db51710439f9f462d1fc98f698eca7ce9777#keys-1> <https://w3id.org/security#publicKeyHex> "03848af62bffceb57631780ac0e0726106ee1c23262d6fd7ef906559d68f53a551" .\n\
```
VolkerSchiewe: Why is this line added? you didn't changed the context, did you?
If so, this is definitely a breaking change

Exulansis: Previously this bit wasn't tested correctly. If you look at how `normalizedDidDocument` is used on develop, you'll see it was only as output for the stubbed `canonize` mock, so we were expecting a mocked function to return the output it was mocked with more or less.

chunningham: the stubbing of the normalization was removed from the did doc tests, so the test actually expands the context now. the publicKeyHex line should have appeared in earlier tests as it is present in both the context and the didDoc

PR #377, opened 2019-11-25
Release/v3.1.0

VolkerSchiewe:
Looks like there are breaking changes in develop :/

This looks like a breaking one: #371 (custom expiry dates)
This as well: #323 (SoftwareKeyProvider with IVs)


VolkerSchiewe:
That would be a major update because SoftwareKeyProvider has breaking changes.

PR #378, opened 2019-11-28
Remove unneeded test dependencies

VolkerSchiewe:


Exulansis:
Should web3 be removed from `package.json` as well? I can't find any other usage.

PR #398, opened 2020-05-05
397/issuance aud

### LINKED ISSUE #397: Add optional "aud" to credential offer tokens
**Describe the solution you'd like**
A mechanism to describe targeted issuance, where the reciever is known beforhand

chunningham:
closes #397, adds "aud" as an option for all request messages and ensures that a response to a request with an audience is issued by that DID in validateJWT


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jolocom-lib",
-  "version": "4.1.0",
+  "version": "4.2.0",
```
Exulansis, line 3: This should not be part of the PR.

Exulansis:
Seems like #399 has to be merged first, since it introduces changes to the Identity Wallet class.

Exulansis:
The `validateJWT` function will need to be broken down, and the underlying validation logic needs to be formalized. A separate issue can be created to address this.

PR #403, opened 2020-06-10
fix(enc): enc/dec in/out use base64

chunningham:
encode the result of encrypting and request string for decrypting as base64


mnzaki:
Should this wait till we've implement actual asymmetric encryption instead of whatever eccrypto is doing? cc @chunningham

chunningham:
superceded by #411

Issue #409, opened 2020-06-23
expose signing on IdentityWallet


chunningham:
Exposing signing functionality on the IdentityWallet class

PR #410, opened 2020-06-23
feat(signing): expose signing on IdentityWallet

chunningham:
addresses #409, note that this method may be unsafe and could probably use some way of securing it.


Exulansis:
Implemented in #424

PR #413, opened 2020-07-16
Add Interaction token type constant inside the JWT message payload

mnzaki:
This simplifies life in SDK internals, at the lower ("purer") layers of the `InteractionManager` after the JWT is peeled away.
The the message ("JWT payload") lacks any type information, creating a need to pass the type around along with the actual message ("token"), and also creating an unnecessary need for typescript guards and what not.

This is not a full solution yet though, but a first step

### DIFF
--- a/ts/identityWallet/identityWallet.ts
+++ b/ts/identityWallet/identityWallet.ts
@@ -4,7 +4,7 @@ import { SignedCredential } from '../credentials/signedCredential/signedCredenti
 import { ExclusivePartial, IIdentityWalletCreateArgs } from './types'
 import { Identity } from '../identity/identity'
 import { JSONWebToken } from '../interactionTokens/JSONWebToken'
-import { InteractionType } from '../interactionTokens/types'
+import { InteractionType, IInteractionToken } from '../interactionTokens/types'
 import { PaymentResponse } from '../interactionTokens/paymentResponse'
 import { PaymentRequest } from '../interactionTokens/paymentRequest'
 import { Authentication } from '../interactionTokens/authentication'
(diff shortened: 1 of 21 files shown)


ts/identityWallet/identityWallet.ts
```diff
@@ -4,7 +4,7 @@ import { SignedCredential } from '../credentials/signedCredential/signedCredenti
 import { ExclusivePartial, IIdentityWalletCreateArgs } from './types'
 import { Identity } from '../identity/identity'
 import { JSONWebToken } from '../interactionTokens/JSONWebToken'
-import { InteractionType } from '../interactionTokens/types'
+import { InteractionType, IInteractionToken } from '../interactionTokens/types'
```
chunningham, line 7: The pattern of these being in the jolocom-proto-ts module obscures the changes made and makes lib pr's quite hard to review, I think we should reconsider it

ts/interactionTokens/JSONWebToken.ts
```diff
@@ -69,7 +69,7 @@ const convertPayload = <T>(args: TransformArgs) => ({
 /* Generic class encoding and decodes various interaction tokens as and from JSON web tokens */
 
 @Exclude()
-export class JSONWebToken<T> implements IDigestable {
+export class JSONWebToken<T extends IInteractionToken> implements IDigestable {
```
chunningham, line 72: I think I dont like this, as we're now enforcing some content in the body, which may conflict with established standards and make moving towards interop more annoying. examples:

1. the [DIF Presentation exchange](https://identity.foundation/presentation-exchange/#presentation-definition) uses a "type" field in it's payload, which this PR is not compatible with. We could wrap in another object in the body but that is unnecessary imo

2. the [JWM spec](https://www.ietf.org/archive/id/draft-looker-jwm-01.txt) defines the message type explicitly outside the payload. If we plan to adopt this in future we'd simply be reverting this change, and perhaps we should take inspiration from JWM if we're already making breaking changes

PR #422, opened 2020-08-13
fix(docs): typo in interactionFlows

PeterTheOne:


PeterTheOne:
One of those is also in https://github.com/jolocom/jolocom-lib/pull/420

Exulansis:
Thank you for reporting, this issue has been handled in #435.

PR #425, opened 2020-08-27
feat(slip0010 recovery): basic support for deriving ed25519 keys usig slip0010

Exulansis:
Also, by extension, adds support for recovering a local identity from a seed.

Currently blocked by a new release of the `native-utils`.


tests/didMethods/jolo/joloRegistrar.test.ts
```diff
@@ -0,0 +1,82 @@
+import * as sinon from 'sinon'
```
Exulansis, line 1: Will bring back before merge

Exulansis:
Also brings in tests for `parseAndValidate`, and for `JolocomRegistrar`.

Fixes a bug in `parseAndValidate.interactionToken` which would lead to signatures not verifying correctly.

PR #433, opened 2020-10-05
Populate JWT header ALG correctly

mnzaki:


chunningham (approved):
looks good to me :+1:

PR #436, opened 2020-10-26
Add to/from JSON conversion to Identity

clauxx:


Exulansis (approved):
Looks fine.
I think can also skip using class-transformer in this case, and simply set / extract the class properties manually (given how straight forward the parsing / serialization is.)

PR #441, opened 2020-11-11
allow jolo-like did methods with a different prefix

chunningham:
allows for DID methods with the same architecture as `did:jolo` but a different anchoring network to easily integrate with the library


Exulansis (changes requested):
Looks good! 
This could use a test or two to ensure that the correct DID prefix is propagated (e.g. across created DID Documents, Signed Credentials, as part of recovery). Especially to ensure we don't accidentally break this functionality later.

PR #452, opened 2021-01-27
Update @jolocom-protocol-ts and replace exported type definitions

mnzaki:
This updates to a version of the protocol that includes `CredentialOffer2` based on https://github.com/jolocom/jolocom-protocol-ts/pull/7


chunningham (approved):
nothing wrong here, but why move diddoc out?

mnzaki:
As discussed on standup, DID Doc attribute types were moved for consistency with the other types being moved.
The fact that it's in this specific PR is because they were also updated in jolocom-protocol-ts, so keeping things consistent

PR #471, opened 2022-05-10
Ag2 321/update offer types

Exulansis:
Mostly protocol-ts updates to support more robust offer styles.


clauxx (approved):
This is a breaking change, so might as well make it 6.0.0

